PR #6, opened 2024-02-07
Implement ETL process for migrating Schemes & Concepts from RDM tables to Lingo models

### LINKED ISSUE #2: Script the ETL of data from the existing RDM tables into the new resource models
When users upgrade to version 8 and want to migrate to the new RDM, they will need a mechanism to "ETL" their data from the existing RDM tables into the new RDM resource models.

We should build this script in a way that it can initially be used as a CLI (ie Django management command or similar), but ultimately we should be able to reuse the code underlying this command to build a front end around this as a UI bulk data operation.  for this reason, it may be best to use the staging tables that Arches already supplies in this effort.

As part of this effort, we will need to definitively map each data element in the existing RDM tables to the new models.  in some cases, data from existing tables (eg `values`) will need to go into many models, so this will not be a simple mapping of columns to nodes in models.

To facilitate an initial MVP here, we should use the RDM models as provided and modify them where needed to simplify data entry.  For example, resource instances nodes referencing the "RDM system" models might be changed for the time being to use a simpler/more convenient datatype such as "string".  The initial goal being to load the data into the model shapes as simply as possible so that we can start to see data loaded into these models at scale and begin work on some of the interfaces that only dependent on a indexed representation of those data (search, some APIs, etc).

johnatawnclementawn:
### Types of changes

-   [ ] Bugfix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description of Change

Uses the Arches ETL manager to extract & transform schemes and concepts from an existing Arches Reference Data Manager schema into the Arches Lingo `Scheme` and `Concept` Models.

### Issues Solved

Resolves #2

### Checklist

-   [ ] Unit tests pass locally with my changes
-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I have added necessary documentation (if appropriate)

#### Ticket Background
*   Sponsored by: 
*   Found by: @ 
*   Tested by: @ 
*   Designed by: @


arches_rdm/etl_modules/migrate_rdm.py
```diff
@@ -0,0 +1,195 @@
...
+                    resourceid=concept_to_load["resourceinstanceid"],
+                    tileid=tile_id,
+                    parenttileid=parent_tile_id,
+                    value=tile_value_json,
+                    nodegroup_depth=nodegroup_depth,
+                    source_description="{0}: {1}".format(concept_to_load["type"], nodegroup_alias),  # source_description
+                    passes_validation=passes_validation,
+                    operation=operation,
+                ))
+        staged_tiles = LoadStaging.objects.bulk_create(tiles_to_load)
```
jacobtylerwalls, line 88: Nice! This bulk create is working well. My jobs failed after this point, but this part seems to work quickly enough that you don't need celery.

arches_rdm/templates/views/components/etl_modules/migrate-rdm.htm
```diff
@@ -0,0 +1,31 @@
+{% load i18n %}
+
+{% block task_details %}
+<!-- ko ifnot: loading() -->
```
jacobtylerwalls, line 4: It seems this task details block is leaking into the task status area:

Screenshot 2024-02-08 at 6 12 48 PM

jacobtylerwalls (commented):
Nice progress!

arches_rdm/etl_modules/migrate_rdm.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+        message = "load event created"
+        return {"success": True, "data": message}
+
+    def write(self, request):
+        self.loadid = request.POST.get("loadid")
+        if models.Concept.objects.count() > 500:
+            response = self.run_load_task(self.userid, self.loadid)
+        else: 
+            response = self.run_load_task_async(request, self.loadid)
```
jacobtylerwalls, line 231: If and else seem like they're swapped -- would expect async for higher volume of concepts

arches_rdm/migrations/0002_rdm_etl_module.py
```diff
@@ -0,0 +1,56 @@
...
+            'RDMMigrator',
+            '{"bgColor": "#ffa564", "circleColor": "#ffd2b1", "show": true}',
+            'fa fa-usb',
+            'migrate-rdm',
+            6,
+            'migrate-rdm-help',
+            true);
+        """
+    remove_migrate_rdm = """
+        DELETE FROM load_staging WHERE loadid IN (SELECT loadid FROM load_event WHERE etl_module_id = '11cad3ca-e155-44b1-9910-c50b3def47f6');
```
jacobtylerwalls, line 46: Likely needs a `DELETE FROM load_errors` also

johnatawnclementawn: Good 🎣

robgaston (approved):

PR #10, opened 2024-04-11
updates application structure

chrabyrd:


README.txt
robgaston, line 1: @chrabyrd we should replace this with content that outlines project setup, as per https://github.com/archesproject/arches_rdm_example_project/blob/main/README.md

yarn.lock
robgaston, line 1: @chrabyrd as we are moving to npm, shouldn't this be replaced with `package-lock.json`?

robgaston (changes requested):
@chrabyrd this is looking very close, but we should look at updating the `README.md` and moving from yarn to npm.  thanks!

chrabyrd:
> @chrabyrd this is looking very close, but we should look at updating the `README.md` and moving from yarn to npm. thanks!

Ah yes -- this was done before the cutover from yarn to npm. Will update! 👍

robgaston (approved):

PR #27, opened 2024-06-10
Fixes lingo package for new structure

robgaston:


johnatawnclementawn (approved):
lgtm!

I added the missing concepts/collections that George created as placeholders for the models. This shouldn't matter much as we progress towards swapping those nodes out for reference datatype nodes.

PR #30, opened 2024-06-12
Adds pre-commit hooks

johnatawnclementawn:


johnatawnclementawn:
Hey @chrabyrd - @jacobtylerwalls and I were stubbing our toes on [failing tests](https://github.com/archesproject/arches-lingo/actions/runs/9484382844/job/26133739372). It looks like CI is caching the dev dependencies which were being changed here.
It seems like there are a handful of options - one is to point to a specific commit hash on the dev-dependencies branch (as I've done here). The other could be to add an `npm cache clean` to the GH action, but we tested that locally and it didn't seem to help.

chrabyrd:
> Hey @chrabyrd - @jacobtylerwalls and I were stubbing our toes on [failing tests](https://github.com/archesproject/arches-lingo/actions/runs/9484382844/job/26133739372). It looks like CI is caching the dev dependencies which were being changed here (not picking up the [addition](d75232c50957237421aece034456b785ad05bb85) of the helper that allows prettier and eslint to play nicely). It seems like there are a handful of options - one is to point to a specific commit hash on the dev-dependencies branch (as I've done here). The other could be to add an `npm cache clean` to the GH action, but we tested that locally and it didn't seem to help.

@johnatawnclementawn could this be a `package-lock.json` issue? I don't love the idea of de-standardizing `package.json` just to make a test runner pass. I'd rather find a solution instead of a workaround, that way if this happens with developers out in the wild we have an answer for them.

package.json
```diff
@@ -10,23 +10,27 @@
...
+        "prettier:check": "prettier arches_lingo/src --check",
+        "prettier:fix": "prettier arches_lingo/src --write",
         "ts:check": "vue-tsc --noEmit",
         "ts:watch": "vue-tsc --watch --noEmit",
         "start": "cross-env NODE_OPTIONS=--max-old-space-size=2048 webpack serve --config ./webpack/webpack.config.dev.js",
         "vitest": "vitest --run --coverage"
     },
     "devDependencies": {
-        "arches-dev-dependencies": "archesproject/arches-dev-dependencies#dev/7.6.x"
+        "arches-dev-dependencies": "archesproject/arches-dev-dependencies#d75232c50957237421aece034456b785ad05bb85"
```
chrabyrd, line 21: I believe it would be better to find a solution instead of a workaround here. I've noticed `package-lock.json` hasn't been updated -- could that be the culprit?

chrabyrd (changes requested):

package.json
johnatawnclementawn: Apologies - I didn't commit `package-lock.json`, but it was updated when I used the direct commit hash. When reverting back to `dev/7.6.x` I experience the same issue. It seems that npm isn't pulling the latest from the dev-deps.

I totally agree that it's suboptimal to de-standardize `package.json`...

johnatawnclementawn: :facepalm: fixed by removing & recreating package-lock.json

chrabyrd (approved):
lgtm!

Issue #29, opened 2024-06-12
Fix RDM to Lingo migration path re. updated models


johnatawnclementawn:
The migration path was broken during latest modeling effort. Need to remap RDM values to nodes

johnatawnclementawn:
Resolved in #33

PR #33, opened 2024-07-01
#29 Remap RDM to Lingo ETL

johnatawnclementawn:
- resource descriptors were broken (pointing to removed nodes), fixed on Concept & Scheme graphs
- Language and Scheme Identifier Type collections were missing values that stood in as placeholders until we can switch over to reference datatype
- Fixed migration path to reflect removed/replaced nodes (e.g. name_content -> appellative_status_ascribed_name_content)


robgaston (approved):
this looks good to me!  +1 for moving constants

PR #35, opened 2024-07-17
Updates Lingo to latest Arches Application pattern in v7.6.x

johnatawnclementawn:
Re. [#11016](https://github.com/archesproject/arches/pull/11016) - updates Arches-Lingo application to follow new Arches Apps paradigm


arches_lingo/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class Arches_LingoConfig(AppConfig):
+    name = "arches_lingo"
+    verbose_name = "Arches_Lingo"
```
jacobtylerwalls, line 6: This should be a human-readable name. For that reason, we've removed it from the template, but may as well pick one here and use it. Shows up in the django admin, I think.

```suggestion
    verbose_name = "Arches Lingo"
```

jacobtylerwalls (changes requested):
Couple nits:
- updateproject still leaves a small gitignore change
- we need the changes from archesproject/arches#11252
  -- If you're wondering why updateproject doesn't do that, it's because we don't just nuke people's github workflows if they had any already

johnatawnclementawn:
> * we need the changes from [fixes GH test runner branch pollution #11251 arches#11252](https://github.com/archesproject/arches/pull/11252)
>   -- If you're wondering why updateproject doesn't do that, it's because we don't just nuke people's github workflows if they had any already

Yeah, I was going to ask about that... that seems reasonable. Would it be worth putting something in the release notes to prompt folks to reconcile their `main.yml` so they don't have to make a bunch of test commits like me?

jacobtylerwalls:
I may be fully/partially wrong on that -- worth a look at updateproject to see what it actually does. I think the app template is being fixed as we speak, see https://github.com/archesproject/arches/pull/11252#issuecomment-2259197262

jacobtylerwalls (approved):

Issue #36, opened 2024-07-31
Add Arches-References dependency


johnatawnclementawn:
Now that the controlled list manager and reference datatype have been broken out into the `Arches-References` application, we should add that as a dependency in Arches-Lingo

johnatawnclementawn:
Resolved in #37

PR #37, opened 2024-07-31
#36 Add arches references as installed app

johnatawnclementawn:
~~Currently dependent on https://github.com/archesproject/arches-lingo/tree/updates_to_latest_v76x~~


jacobtylerwalls (approved):
Great!

We'll debug the failure re: missing tsconfig.json from arches-references later.

Workaround locally is to use `-e` to install things.

PR #38, opened 2024-08-02
Implement new base-root.htm template

### LINKED ISSUE #3: Create new RDM plugin as a Vue component/application
We need this application to define and add a new plugin which will be the entrypoint for the new RDM UI.

This plugin should be built using Vue, so the root component should be essentially a Vue application.

targeting the simple mockup below, we should build dummy components to create the layout for this initial UI:

Screenshot 2023-12-12 at 12 00 59 PM

jacobtylerwalls:
- manage.py setup_db for good measure to remove plugin
- visit web root
- result: hello world, no arches chrome

Closes #3


chrabyrd:
👀

chrabyrd:
:wave: @jacobtylerwalls I'm bouncing this back to you, the tsconfig/frontend stuff is fixed but now the tests are failing with:

```
RuntimeError: Model class arches_references.models.List doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

jacobtylerwalls:
Great! Test is now just failing on something substantive, looks like an out-of-date graph constant. I'll ask @johnatawnclementawn to take a look in #40, suggest merging this anyway to unblock things.

chrabyrd (approved):
lgtm!

arches_lingo/settings.py
```diff
@@ -145,6 +144,7 @@
     "corsheaders",
     "oauth2_provider",
     "django_celery_results",
+    "arches_references",
```
jacobtylerwalls, line 147: Don't we want this above arches since it's an arches application?

chrabyrd: it doesn't matter, as long as `arches.app` is at the bottom

jacobtylerwalls: I thought that was only true for templates, not css/img discovery.

chrabyrd: just testing with an image and the results are as you'd expect ( project -> apps -> core )

jacobtylerwalls: Okay, I'm with you now. I forgot how specialized we made `ArchesApplicationsStaticFilesFinder`, which even if there's an `"arches"` in the middle of INSTALLED_APPS, there won't be anything under `arches/media`, so this should be A-OK.

johnatawnclementawn: @jacobtylerwalls @chrabyrd back from the future - `arches_references` has to go above `arches` because there are commands in `arches_references/packages.py` that need to be included in the packages options.  I'll fix as part of #53

PR #39, opened 2024-08-05
Add login interface #13

### LINKED ISSUE #13: Implement token-based user authentication front end
This will likely leverage the Django REST Framework (DRF).

A component should be added to the header. It should:
- Have a `register` and `login` button
- Display the name of the user when logged in
- Have a `logout` button when the user is logged in
- Toast service for login failure/success

For this iteration the `register` button should redirect to the Arches user registration page. However the `login` button should display a component that accepts a username and password, and attempts to login the user via DRF API call. It should also contain a link for logging in via Arches login page to support SSO/2FA users.

blocked by: https://github.com/archesproject/arches/issues/11261 and #12

jacobtylerwalls:
Closes #13 

Depends on archesproject/arches#11284 and https://github.com/archesproject/arches/pull/11331

Styling will be finalized later.

### Testing instructions
Visit the site (at root), ensure you don't have a csrf cookie (first-time visitor).
  - [ ] Login works in success path
  - [ ] Appropriate message shown in failure path
  - [ ] Logout works
  - [ ] Links to register, existing auth form work
  - [ ] Redirect from auth form doesn't go to login page
  - [ ] Refresh doesn't go to login page
  - [ ] Superuser can navigate directly to django admin


tests/tests.py
```diff
@@ -135,7 +135,8 @@ def setUpTestData(cls):
         cls.concepts = [
             ResourceInstance(graph_id=SCHEMES_GRAPH_ID) for _ in range(CONCEPT_COUNT)
         ]
-        ResourceInstance.objects.bulk_create(cls.concepts)
```
chrabyrd, line 138: If this is what's breaking tests on `main`, could you make this into its own PR?

arches_lingo/templates/arches_lingo/root.htm
```diff
@@ -5,5 +5,5 @@
 {% endblock title %}
 
 {% block body %}
-<div id="lingo-mounting-point"></div>
+<div id="lingo-mounting-point" style="font-family: sans-serif;"></div>
```
chrabyrd, line 8: Can this be moved a level lower?

arches_lingo/src/arches_lingo/api.ts
```diff
@@ -0,0 +1,43 @@
+import arches from "arches";
+import Cookies from "js-cookie";
+
+function getToken() {
+    const token = Cookies.get("csrftoken");
+    if (!token) {
+        throw new Error("Missing csrftoken");
```
chrabyrd, line 7: needs i18n

arches_lingo/src/arches_lingo/components/login/LoginPage.vue
chrabyrd, line 1: Does this belong under `components`? IMO it's a top-level thing

chrabyrd (commented):
looks good! Just a few Qs 👍

arches_lingo/src/arches_lingo/components/login/LoginPage.vue
jacobtylerwalls: What do you think about the App.vue component being the only top-level component outside components? As currently implemented, it uses `<component>` to switch between HomePage and LoginPage, but if you don't like that pattern, we can have the LoginPage be the top level component instead of App.vue.

arches_lingo/src/arches_lingo/api.ts
jacobtylerwalls: The only value this has for an end user is to copy/paste it into a support ticket. I don't know that translating improves anything in that regard. The call site catch uses a translated message for the toast summary, and then passes this through untranslated as the detailed message. I guess we have other examples of marking highly technical strings for translation, but I'm not sure asking a translator to translate "Missing csrftoken" provides value. It's also a tricky to implement since we're outside a vue component.

chrabyrd: no that's fair 👍

arches_lingo/src/arches_lingo/components/login/LoginPage.vue
chrabyrd: Ah -- fun discussion abounds 😄 . This is an interesting topic for me.

As I understand it you're suggesting:

```
.
└── arches-lingo/
    └── arches_lingo/
        └── src/
            └── arches_lingo/
                ├── App.vue
                └── components/
                    └── login/
                        ├── LoginPage.vue
                        ├── LoginLinks.vue
                        └── LoginForm.vue
```

which is _fine_. The only problem I have with it is that it combines more-basic stuff with more-complex stuff. I guess unofficially I've been trying to steer us towards atomic design, and this violates that to some degree. Playing it out I could see:

```
.
└── arches-lingo/
    └── arches_lingo/
        └── src/
            └── arches_lingo/
                ├── App.vue
                └── components/
                    ├── login/
                    │   ├── LoginPage.vue
                    │   ├── LoginLinks.vue
                    │   └── LoginForm.vue
                    ├── foo/
                    │   ├── FooStuff.vue
                    │   └── ThingThatUsesFooStuff.vue
                    ├── bar/
                    │   ├── BarComponents.vue
                    │   ├── BarForm.vue
                    │   └── ThingThatUsesBarForm.vue
                    └── baz/
                        └── ThingThatUsesFooStuffAndBarForm.vue ?
```

where `baz` is a complex thing (eg page) that still lives in the `components` directory. Maybe it's just me being pedantic ( 😅 ) but a page shouldn't live among its componentry.

Are you open to something like `views` or `pages`?

jacobtylerwalls: > Are you open to something like views or pages?

Definitely, I'm just flying a little blind as I don't know how many pages or views we're going to have. More than one?

jacobtylerwalls: 0c06de0b6e29c227299ff7665234a502337cbf9c

chrabyrd (approved):
lgtm 👍

PR #51, opened 2024-08-13
Fix TS pre-commit hook

jacobtylerwalls:
Partner to archesproject/arches#11250.

To test, merge this branch into #39, and try commenting out the `if (!token) { ...` block of `getToken()` in api.ts. You shouldn't be able to commit that change. Without these changes, the commit goes through because the hook wasn't firing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 ---
 repos:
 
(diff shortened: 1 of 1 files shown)


johnatawnclementawn (approved):
Looks good! 👍🏼  Thanks for the repro instructions

PR #52, opened 2024-08-13
Implement front-end routes #11

### LINKED ISSUE #11: Implement Vue routing
We need to integrate Vue Router into the plugin to enable easy frontend navigation with browser history.

- Setup Vue Router
- Create and link routes to placeholder components
- Implement route lazy loading

There should be `Hello world`-level placeholder components for the welcome page, the user homepage, the search page, the advanced search page, and the Scheme list page.

An assumption with the time estimation of this ticket is that the pattern has been proven out in the Controlled List Manager and will largely be copy/pasted.

jacobtylerwalls:
- Adds frontend routes and hello-world components for them to render to
- Guards all but the login route behind a redirect to the login page
- Forwards the user to the originally requested route post-auth
- Adds lazy routes

Closes #11 

PR currently based off #39


arches_lingo/src/arches_lingo/constants.ts
```diff
@@ -1,7 +1,47 @@
...
+    },
+    {
+        path: "/schemes",
+        name: "schemes",
+        component: () =>
+            import("@/arches_lingo/pages/scheme-list/SchemeList.vue"),
+    },
+];
+
+export const routeNames = {
```
chrabyrd, line 41: should also be moved to whichever file `routes` will live, if they're grouped as `routeNames`. As individual strings yeah here is ok

package.json
```diff
@@ -21,10 +21,10 @@
         "arches-dev-dependencies": "archesproject/arches-dev-dependencies#dev/8.0.x"
     },
     "dependencies": {
-        "arches": "archesproject/arches#dev/8.0.x"
-    },
-    "nodeModulesPaths": {
+        "arches": "archesproject/arches#dev/8.0.x",
+        "vue-router": "^4.4.3"
```
chrabyrd, line 25: I don't feel very strongly about this yet, but I could see `vue-router` going into `dev/8.0.x`

chrabyrd (changes requested):
All in all looks good, just a few minor things. 

Also wanted to touch base on folder hierarchy ( yet again 😄  ). IMO `components` should live outside of `pages`. Happy to chat about it though

package.json
jacobtylerwalls: For sure, let's cross that bridge when we need to.

chrabyrd (approved):
lgtm 👍

Issue #14, opened 2024-05-01
Create a side navigation bar


chrabyrd:
This should exist outside of the Vue router. For this iteration it should only have basic styling, and reflect which page the user on via icon highlight or button pressed state -- depending on implementation.

It should contain links for:
- User homepage
- Search page
- Scheme list page

It should be responsive, and support becoming a topnav or integrating into the current topnav on mobile devices.

jacobtylerwalls:
@chrabyrd is there a mockup?

chrabyrd:
resolved in #62

PR #62, opened 2024-08-16
create sidenav #14

chrabyrd:


arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
```diff
@@ -0,0 +1,56 @@
...
+
+<template>
+    <aside class="sidenav">
+        <div
+            v-for="item in items"
+            :key="item.name"
+        >
+            <RouterLink :to="{ name: item.name }">
+                <Button>
+                    <i :class="item.icon"></i>
```
jacobtylerwalls, line 28: Let's add an aria-label. Or maybe better, a tooltip? Or a follow-up issue?

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -38,10 +39,21 @@ router.beforeEach(async (to) => {
...
-        <div style="margin: 1rem">
-            <RouterView />
+    <main
+        style="
+            font-family: sans-serif;
+            height: 100vh;
+            width: 100vw;
+            display: flex;
+            flex-direction: column;
+        "
```
jacobtylerwalls, line 49: Without this in root.htm:
```
{% block body_attributes %}style="margin: 0"{% endblock %}
```

I get 8px of margin, causing scrolls (because I'm a pest and I configure MacOS to have Show Scroll bars > Always):



margin

jacobtylerwalls (approved):
Looks great, two small nits you can punt if you like.

arches_lingo/src/arches_lingo/App.vue
chrabyrd: OH! this is fixed by https://github.com/archesproject/arches/pull/11343. I've added it manually to my local and have been developing against it 👍

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
chrabyrd: Done and done

arches_lingo/src/arches_lingo/App.vue
jacobtylerwalls: Hmm, I tried that and it didn't help but I'll check again tomorrow maybe I botched the build.

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
jacobtylerwalls: Whoops, I guess I shouldn't have suggested one or the other, I think we still need accessibility features on the button/icon combo, probably aria-label on the button and aria-hidden=True on the icon, something like that. I don't think the tooltip is enough here, do you agree @johnatawnclementawn?

chrabyrd: no worries! Done and done 👍

jacobtylerwalls (approved):
great! 👌

arches_lingo/src/arches_lingo/App.vue
jacobtylerwalls: no need to wait for me, let's merge 👌

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
johnatawnclementawn: `aria-hidden` on the icon is good, but I'd caution against using a button inside of a link. When I traverse through this using VO, `Home` and `Search` are read out twice first as a link, then as a button. The expected interaction here are navigation to a different page, so best to just use a link there.

chrabyrd: @johnatawnclementawn this is a good catch!

Issue #56, opened 2024-08-15
Application does not retain user logged-in status


chrabyrd:
To reproduce:

1. visit the root site ( eg localhost:8000/ )
2. login
3. refresh the page
4. notice that instead of the page the user is redirected to after login, the login form is displayed

jacobtylerwalls:
Thanks for finding, I must have broken this after adding the unauthorized user redirect.

chrabyrd:
resolved in #64

Issue #61, opened 2024-08-16
same-named routes conflict


chrabyrd:
To reproduce:

1. login to the application
2. go to the search page via header link ( or homepage link )
3. refresh the application
4. notice that the Arches search page is rendered

chrabyrd:
resolved in #64

Issue #57, opened 2024-08-15
Refreshing the login page after logging in results in url becoming un-navigatable  


chrabyrd:
To reproduce:

1. navigate to the application root ( eg localhost:8000 )
2. login
3. refresh the page
4. notice that instead of the page the user is redirected to after login, the login form is displayed
5. notice the url now has `/login/root` appended
6. refresh the page again
7. get `Page not found` error

chrabyrd:
resolved in #64

PR #64, opened 2024-08-19
fixes user retention #56

chrabyrd:
This resolves: #56, #57, and #61.

It also raises interesting questions about session management and where user auth should live. Right now we're fetching the user for each page load. Not a huge overhead, and it gives the added bonus on knowing when a user has logged out via arches. Did another quick read re Django sessions and I'm wondering if there's a better way.

Also, right now as a placeholder for more complex logic we're manually checking if a user is `anonymous` or not. We're also defining whether a route should be "protected" in `routes.ts` . What about more complex logic, like if a page ( or component ) should only be rendered if a user belongs to a certain permission group? Moving the permission check to the page level feels more explicit, but possibly less DRY.


arches_lingo/src/arches_lingo/api.ts
```diff
@@ -16,11 +16,11 @@ export const login = async (username: string, password: string) => {
         body: JSON.stringify({ username, password }),
     });
     try {
-        const parsed = await response.json();
+        const responseJson = await response.json();
```
chrabyrd, line 19: these names got changed as I was exploring earlier solutions with attempting to access session cookies. Happy to revert if need be /shrug

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -22,32 +27,36 @@ provide(userKey, { user, setUser });
 
 const router = useRouter();
 const route = useRoute();
+const toast = useToast();
+const { $gettext } = useGettext();
 
-const isAuthenticated = computed(() => {
-    return user.value && user.value.username !== "anonymous";
-});
+router.beforeEach(async (to, _from, next) => {
```
chrabyrd, line 33: This pattern no longer tacks on the `next` route for unpermissioned user login redirects ( eg `localhost:8000/login/search` )

jacobtylerwalls: Much better solution, thanks.

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -22,32 +27,36 @@ provide(userKey, { user, setUser });
...
-        to.name !== routeNames.login
-    ) {
-        return { name: routeNames.login, params: { next: to.name as string } };
+        const requiresAuth = to.matched.some(
+            (record) => record.meta.requiresAuth,
+        );
+        if (requiresAuth && userData.username === "anonymous") {
+            throw new Error();
+        } else {
+            next();
```
jacobtylerwalls, line 44: nice!

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -22,32 +27,36 @@ provide(userKey, { user, setUser });
...
+const toast = useToast();
+const { $gettext } = useGettext();
 
-const isAuthenticated = computed(() => {
-    return user.value && user.value.username !== "anonymous";
-});
+router.beforeEach(async (to, _from, next) => {
+    try {
+        const userData = await fetchUser();
+        setUser(userData);
```
jacobtylerwalls, line 36: Great solution, let's just only fetch the data if it's missing -- my idea was to fetch user details on refresh, not on in-app route-navigation.

```suggestion
        let userData = user.value;
        if (!userData) {
            userData = await fetchUser();
            setUser(userData);
        }
```

jacobtylerwalls (changes requested):

jacobtylerwalls:
> What about more complex logic, like if a page ( or component ) should only be rendered if a user belongs to a certain permission group?

Let's not get too fancy with the routing: the private routing is just there to ensure people get to the login page when they need to: it's the API that will enforce that permissioned users only see the data they can access.

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -22,32 +27,36 @@ provide(userKey, { user, setUser });
...
+        } else {
+            next();
+        }
+    } catch (error) {
+        toast.add({
+            severity: ERROR,
+            life: DEFAULT_ERROR_TOAST_LIFE,
+            summary: $gettext("Login required."),
+            detail: error instanceof Error ? error.message : undefined,
+        });
```
jacobtylerwalls, line 52: I like this in principle, but it's too much when just navigating to the home page and getting redirected to login.

arches_lingo/src/arches_lingo/App.vue
chrabyrd: ++

arches_lingo/src/arches_lingo/App.vue
chrabyrd: cool cool, removed the `requiresAuthentication` req for the home route 👍

chrabyrd:
There's some not-great UX around the sign out button being visible to a not-logged-in user on the homepage now. IMO punt since this is a placeholder interaction

jacobtylerwalls (changes requested):
Ah, I don't think we can just set requireAuth false on the root, because now when I navigate to the root as a logged out/anonymous user, I bypass login and go straight to the home page. I guess I'd rather just suppress the toast somehow.

chrabyrd:
> Ah, I don't think we can just set requireAuth false on the root, because now when I navigate to the root as a logged out/anonymous user, I bypass login and go straight to the home page. I guess I'd rather just suppress the toast somehow.

Hmm I'm a little confused here,  the user should get redirected to the login page when navigating to root, but shouldn't get a toast about it?

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -18,36 +25,49 @@ const user = ref<User | null>(null);
...
-router.beforeEach(async (to) => {
-    if (
-        !isAuthenticated.value &&
-        // Avoid an infinite redirect
-        to.name !== routeNames.login
-    ) {
-        return { name: routeNames.login, params: { next: to.name as string } };
+        if (!userData) {
+            userData = await fetchUser();
+            setUser(userData);
```
jacobtylerwalls, line 41: Apologies if my last suggested change is what set this awry, but I'm now noticing that when I:
- start with a fresh browser (no cookies)
- navigate to root (result: I get a csrftoken cookie as expected)
- I log in
- nothing happens (network tab shows login succeeded, but i'm not navigated away. now the session cookie exists, though)

Maybe we need to check for the existence of the session cookie also before we short-circuit like this?

jacobtylerwalls: ...or check for auth'd user (not anonymous user)?

chrabyrd: no worries and good catch!

jacobtylerwalls (approved):
+1 tested, thanks for the improvements here!

PR #65, opened 2024-08-19
Use valid `FILE_TYPE_CHECKING` value

jacobtylerwalls:
Follow-up to archesproject/arches#11353.
Matches default for new projects.


johnatawnclementawn (approved):
LGTM!

PR #69, opened 2024-08-20
Improve i18n for user greeting

jacobtylerwalls:
Avoid +'ing strings together, which isn't helpful to translators.


chrabyrd (approved):
lgtm!

Issue #70, opened 2024-08-20
User state is stale after logout


jacobtylerwalls:
- log in as user1
- see "hello user1"
- logout
- log in as user 2
- see "hello user1"
- refresh
- see "hello user2"

chrabyrd:
resolved in #71

PR #71, opened 2024-08-20
prevents stale user after logout #70

chrabyrd:


jacobtylerwalls (approved):
Thanks, fixes the issue. There must be a better way to handle the client state so we don't have to fetch so much now, but we can polish that later.

jacobtylerwalls (changes requested):
Actually, I'm showing this is fixed with just:
```diff
diff --git a/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue b/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
index 9361665..def4c3e 100644
--- a/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
+++ b/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
@@ -20,11 +20,12 @@ const { $gettext } = useGettext();
 const toast = useToast();
 const router = useRouter();
 
-const { user } = inject(USER_KEY) as UserRefAndSetter;
+const { user, setUser } = inject(USER_KEY) as UserRefAndSetter;
 
 const issueLogout = async () => {
     try {
         await logout();
+        setUser(null);
         router.push({ name: routeNames.login });
     } catch (error) {
         toast.add({
```

chrabyrd:
> Actually, I'm showing this is fixed with just:
> 
> ```diff
> diff --git a/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue b/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
> index 9361665..def4c3e 100644
> --- a/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
> +++ b/arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
> @@ -20,11 +20,12 @@ const { $gettext } = useGettext();
>  const toast = useToast();
>  const router = useRouter();
>  
> -const { user } = inject(USER_KEY) as UserRefAndSetter;
> +const { user, setUser } = inject(USER_KEY) as UserRefAndSetter;
>  
>  const issueLogout = async () => {
>      try {
>          await logout();
> +        setUser(null);
>          router.push({ name: routeNames.login });
>      } catch (error) {
>          toast.add({
> ```

True, but then we lose the feature of dealing with if a user has logged out in the Arches (not lingo) application. I know requesting the user every page isn't ideal but IMO it's better/safer.

jacobtylerwalls:
Ah, I forgot about that case. Approving!

jacobtylerwalls (approved):

PR #74, opened 2024-08-21
Improve RDM to Lingo migration

johnatawnclementawn:
Resolves #68 

You should be able to review this by:
1. checking out the branch
2. reload package
3. visit bulk data manager (select Migrate to Lingo)
4. select a Lingo thesaurus, e.g. SKOS
5. Resources created should match records in the RDM

PR #76, opened 2024-08-22
Add basic search backend #67

### LINKED ISSUE #67: Implement basic search page backend
This should handle fuzzy-searching, and the data returned should include an item's hierarchical path

jacobtylerwalls:
Add basic search backend at api/search. Closes #67

GET params:
- term: search term
- maxEditDistance (fuzziness)
- page
- items (per page)
- exact

All are optional. If you don't provide a search term then the cached hierarchies are rebuilt. A use case for this might be some sort of "preflight" request to warm the cache when the search page loads and before the user interacts with the search input. All of this is up for discussion. The default cache life is 5 minutes and is configurable.

### Demo
http://127.0.0.1:8000/api/search?term=oss
<details>
<summary>Result</summary>

```
[
  {
    "id": "d4e4db86-a948-4d6a-b89f-d1682d038afe",
    "labels": [
      {
        "language": "de",
        "value": "Knochen",
        "valuetype": "prefLabel"
      },
      {
        "language": "pt",
        "value": "osso",
        "valuetype": "prefLabel"
      },
      {
        "language": "en",
        "value": "bone (material)",
        "valuetype": "prefLabel"
      },
      {
        "language": "fr",
        "value": "os",
        "valuetype": "prefLabel"
      }
    ],
    "parents": [
      {
        "id": "b73e741b-46da-496c-8960-55cc1007bec4",
        "labels": [
          {
            "language": "en-US",
            "value": "AAT Entries",
            "valuetype": "prefLabel"
          }
        ]
      },
      {
        "id": "7764512c-494b-46e5-ad33-223836c8518b",
        "labels": [
          {
            "language": "en-US",
            "value": "Materials",
            "valuetype": "prefLabel"
          }
        ]
      }
    ],
    "polyhierarchical": false
  },
  {
    "id": "07dbe013-7dcf-4dd7-9df1-e72a9a855da5",
    "labels": [
      {
        "language": "fr",
        "value": "bois",
        "valuetype": "prefLabel"
      },
      {
        "language": "en",
        "value": "wood (plant material)",
        "valuetype": "prefLabel"
      }
    ],
    "parents": [
      {
        "id": "b73e741b-46da-496c-8960-55cc1007bec4",
        "labels": [
          {
            "language": "en-US",
            "value": "AAT Entries",
            "valuetype": "prefLabel"
          }
        ]
      },
      {
        "id": "7764512c-494b-46e5-ad33-223836c8518b",
        "labels": [
          {
            "language": "en-US",
            "value": "Materials",
            "valuetype": "prefLabel"
          }
        ]
      }
    ],
    "polyhierarchical": false
  }
]
```
</details>

### Testing instructions
1. Load some data following instructions at #74
2. Checkout this branch, run migration
3. Make some requests
4. You can compare the search results to the haystack by viewing api/concept_trees

### TODO:
- [x] Add python tests


arches_lingo/views/trees.py
```diff
@@ -0,0 +1,365 @@
...
+        self.broader_concepts: dict[str : set[str]] = defaultdict(set)
+        # key=resourceid (str) val=set of scheme resourceids (str)
+        self.schemes_by_top_concept: dict[str : set[str]] = defaultdict(set)
+
+        self.read_from_cache()
+
+        # Not currently cached because written to during serialization.
+        self.polyhierarchical_concepts = set()
+
+    def read_from_cache(self):
```
jacobtylerwalls, line 96: Caching isn't working on my local right now. I'll look into it a little bit, but we may also find we don't even need this. Would be nice to check this on realistic cloud infra with a realistic dataset.

jacobtylerwalls: Culprit was a typo. Turns out you can never be in too much of a hurry to skip creating python constants 😅

arches_lingo/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [("models", "11042_update__arches_staging_to_tile")]
+
+    forward = textwrap.dedent(
+        f"""
+    CREATE MATERIALIZED VIEW arches_lingo AS (
```
jacobtylerwalls, line 23:
```suggestion
    CREATE MATERIALIZED VIEW arches_lingo__vw_label_values AS (
```

arches_lingo/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+        ROW_TO_JSON(JSONB_EACH(t.tiledata -> '{CONCEPT_NAME_CONTENT_NODE}'))
+            -> 'value' ->> 'value' AS value
+    FROM
+        tiles t
+    ORDER BY
+        conceptid
+    );"""
+    )
+
+    reverse = "DROP MATERIALIZED VIEW arches_lingo;"
```
jacobtylerwalls, line 35:
```suggestion
    reverse = "DROP MATERIALIZED VIEW arches_lingo__vw_label_values;"
```

arches_lingo/views/__init__.py
chrabyrd, line 1: Why import here? IMO it's more explicit to import from the file

arches_lingo/views/trees.py
chrabyrd, line 1: Can the filename be changed to something more descriptive?

chrabyrd, line 1: Should this live in `/api/` instead of `/views/`?

arches_lingo/views/trees.py
```diff
@@ -0,0 +1,363 @@
...
+        paginator = Paginator(concept_query, items_per_page)
+        page = paginator.get_page(page_number)
+
+        data = [
+            self.serialize_concept(str(concept_uuid), parents=True, children=False)
+            for concept_uuid in page
+        ]
+
+        # Todo: filter by nodegroup permissions
+        return JSONResponse(data)
```
chrabyrd, line 349: This has been altered in https://github.com/archesproject/arches-lingo/pull/79 to return more data about the query

arches_lingo/views/trees.py
```diff
@@ -0,0 +1,363 @@
...
+            "schemes": [self.serialize_scheme(scheme) for scheme in self.schemes],
+        }
+        # Todo: filter by nodegroup permissions
+        return JSONResponse(data)
+
+
+@method_decorator(
+    group_required("RDM Administrator", raise_exception=True), name="dispatch"
+)
+class ValueSearchView(ConceptTreeView):
```
chrabyrd, line 313: This is returning some odds results that don't seem to match with the query string in any way. I've tried munging about with `elf.default_sensitivity()` but to no fantastic result. Any ideas?

arches_lingo/views/trees.py
```diff
@@ -0,0 +1,363 @@
...
+
+class LevenshteinLessEqual(Func):
+    arity = 3
+    function = "LEVENSHTEIN_LESS_EQUAL"
+
+
+@method_decorator(
+    group_required("RDM Administrator", raise_exception=True), name="dispatch"
+)
+class ConceptTreeView(View):
```
chrabyrd, line 71: pretty big view. can anything be moved from it to a model or util?

chrabyrd (changes requested):
overall looks good, and I appreciate the amount of work that's gone into powering this (new && different) search 👍 

Mostly minor stuff, but happy to touch base about what I'm imagining is a _too_-fuzzy search 😁

arches_lingo/views/trees.py
jacobtylerwalls: We ended up union'ing this with an `__icontains` search.

arches_lingo/querysets.py
chrabyrd, line 1: Can this be moved into a folder somewhere?

arches_lingo/settings.py
```diff
@@ -110,7 +110,9 @@
         "ENGINE": "django.contrib.gis.db.backends.postgis",
         "HOST": "localhost",
         "NAME": "arches_lingo",
-        "OPTIONS": {},
+        "OPTIONS": {
+            "options": "-c cursor_tuple_fraction=1",
```
chrabyrd, line 114: just curious, what's up with this?

arches_lingo/urls.py
```diff
@@ -3,15 +3,17 @@
...
+from arches_lingo.views.root import LingoRootView
+from arches_lingo.views.api.concepts import ConceptTreeView, ValueSearchView
 
 urlpatterns = [
     path("", LingoRootView.as_view(), name="root"),
     path("login", LingoRootView.as_view(), name="login"),
     path("search", LingoRootView.as_view(), name="search"),
     path("advanced-search", LingoRootView.as_view(), name="advanced-search"),
     path("schemes", LingoRootView.as_view(), name="schemes"),
-    path("api/concept_trees", ConceptTreeView.as_view(), name="concept_trees"),
```
chrabyrd, line 14: nit: dash-delineated

arches_lingo/views/api/concepts.py
```diff
@@ -0,0 +1,87 @@
...
+                )
+            except ValueError as ve:
+                return JSONErrorResponse(
+                    title=_("Unable to perform search."),
+                    message=ve.args[0],
+                    status=HTTPStatus.BAD_REQUEST,
+                )
+        else:
+            concept_query = VwLabelValue.objects.all().order_by("concept_id")
+        concept_query = concept_query.values_list("concept_id", flat=True).distinct()
```
chrabyrd, line 58: nit: can this be named something new instead of renaming what was set in the logic above?

arches_lingo/views/api/concepts.py
```diff
@@ -0,0 +1,87 @@
...
+
+        paginator = Paginator(concept_query, items_per_page)
+        page = paginator.get_page(page_number)
+
+        data = []
+        if page:
```
chrabyrd, line 64: nit: instead of control flow based on presence/absence of a pagintor page, can it based on the data fed into the paginator?

arches_lingo/views/root.py
chrabyrd, line 1: not part of this PR but something to consider. IMO we, as a team, should avoid naming schemes like `App.vue`, `root.htm`, `root.py` etc etc. It's not super friendly to the applet model I see in our future. I'll likely make a cleanup ticket sometime but wanted to put this on your radar for discussion beforehand

chrabyrd (changes requested):
overall looks good and works well! just some minor code style stuff really

arches_lingo/settings.py
jacobtylerwalls: It's anticipating https://github.com/archesproject/arches/issues/11382.

I can go ahead and open a PR against the 8.0 project template at this point.

chrabyrd: Interesting!

arches_lingo/views/api/concepts.py
jacobtylerwalls: The ConceptBuilder does queries to build a tree (or gets the cached one). If there are no search results, there's no need to build a tree, we can just quickly return empty. That was my thinking. I'm not exactly sure what you're suggesting as an alternative. I can clarify this to deindent a little bit.

chrabyrd: ya, I've not put a breakpoint to see if this holds up but

`concept_query = concept_query.values_list("concept_id", flat=True).distinct()`

returns something that will either have a length or not? 

Since `page_number` looks like it will always have a value, the only time there shouldn't be a page is when there's not data. Admittedly this is my assumption.

So instead of `if page:`, can it be `if len(concept_query): ( or whatever its new name is )`?

arches_lingo/settings.py
jacobtylerwalls: Yeah, it's good for lingo because when I'm using .iterator() to visit all the broader concept tiles, we know for a fact we need to exhaust all of them to build the correct tree, we won't quit early sometimes.

arches_lingo/concepts.py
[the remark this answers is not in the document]
jacobtylerwalls: I considered it, but I don't want to encourage running these methods in of a context where the annotations they depend on might not have occurred. I don't want to "advertise" this functionality on the model. Don't try to serialize a concept with information about its children and parents unless you've init'ed the object that built the tree. Something like that.

If we had actual relational models all the way down then for sure, this would all be on the model. But it's Tiles.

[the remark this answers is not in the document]
jacobtylerwalls: I did move to a utils folder 👍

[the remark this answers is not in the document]
jacobtylerwalls: We may find ourselves with a relational model at some point and we can organize the code around models in that case.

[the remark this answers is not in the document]
chrabyrd: ++

arches_lingo/views/api/concepts.py
```diff
@@ -0,0 +1,88 @@
...
+                return JSONErrorResponse(
+                    title=_("Unable to perform search."),
+                    message=ve.args[0],
+                    status=HTTPStatus.BAD_REQUEST,
+                )
+        else:
+            concept_query = VwLabelValue.objects.all().order_by("concept_id")
+        concept_ids = concept_query.values_list("concept_id", flat=True).distinct()
+
+        paginator = Paginator(concept_ids, items_per_page)
```
chrabyrd, line 60: That way, this code gets reduced to something like:

```
        if not concept_ids:
            # No results: don't bother building the concept tree.
            return JSONResponse([])

        paginator = Paginator(concept_ids, items_per_page)
        page = paginator.get_page(page_number)

        builder = ConceptBuilder()

        return JSONResponse([
            builder.serialize_concept(str(concept_uuid), parents=True, children=False)
            for concept_uuid in page
        ])
```

arches_lingo/views/api/concepts.py
jacobtylerwalls: Yeah I can signpost better. But we can't evaluate the queryset too early or we'll unnecessarily fetch lots of objects before doing the pagination to slim down the results.

arches_lingo/views/api/concepts.py
jacobtylerwalls: Yeah, as I mentioned above, it would be a pessimization to evaluate the queryset before the paginator slices it. We need delegate that to the paginator.

arches_lingo/views/api/concepts.py
jacobtylerwalls: In case the implicit boolenness of the Page object is too opaque, I'll change it to `if not len(page)` to clarify it's an iterable of objects.

jacobtylerwalls: Actually, it saves one query in the no-data path to check `page.count`, which seems like something I might ask on the Django forum about because `len()` should work the same way. So I'll do that.

jacobtylerwalls: Oops, this should be .count on the paginator, not the page.

chrabyrd (approved):
looks good 👍

jacobtylerwalls:
Thanks for the review!

PR #86, opened 2024-08-29
Integrate hierarchy view into schemes list placeholder #18, add concept and scheme routing #15

### LINKED ISSUE #15: Add scheme and concept pages, dynamic route matching
In this iteration, this page should not be visible in the sidenav, it should be accessed via URL manipulation only. The page itself should be a `Hello world`-level placeholder.

Ensure the page title in the topnav is updated with the scheme/component name.

### LINKED ISSUE #18: Integrate concept hierarchy component
This should be integrated in three places:

- The scheme list page
- The scheme item page
- The concept item page

This will require a button on each page that opens a side panel that displays the hiearchy.

For this iteration, the only functionality expected is that clicking an item will take the user to the page for that scheme/concept. If the user is on the scheme/concept page and the button to display the hierarchy is clicked, it displayed hierarchy should already be opened to the selected scheme/concept

Screenshot 2024-05-01 at 3 09 26 PM

### LINKED ISSUE #114: Determine future of item icon
In #86, a component is copied over from the controlled list manager `<LetterCircle>`.

We have some options:
- reimplement it entirely
- import it from arches-references
- find another design
- double down on it and put in in arches-vue-utils

I'd suggest finding another design:
- whereas a brand like Google might use a "G" for its brand, this isn't a brand, so I'm not sure a letter inside an icon makes much sense for conveying this information.
- the letter will need to be translated, and now there's no guarantee the letters will even be distinct from each other once translated.
- color choices now might have contrast/size constraints if we're coloring text.

Should we examine alternatives?

jacobtylerwalls:
Visit http://127.0.0.1:8000/schemes to see the hierarchy component on the left in a collapsible panel.

Most of the tree functionality is borrowed from the controlled list manager.

Closes #15
Closes #18
Closes #114

For discussion:
- Language switcher not yet implemented
- Some components (e.g. `<PresentationControls>`) are just imported from arches-references instead of reimplemented. Others e.g. `<TreeRow>` are reimplemented. I chose not to namespace them under `arches_references` under `arches_lingo` because I had no higher higher level components from arches_references that needed to resolve to them.
- Styling not finalized
- Once we know what other pages want this hierarchy, we can lift some of the state up (e.g displayed row).


arches_lingo/src/arches_lingo/pages/SchemeList.vue
```diff
@@ -1 +1,61 @@
...
+import type { Labellable } from "@/arches_lingo/types.ts";
+
+const { $gettext } = useGettext();
+
+// TODO: lift some of this state up?
+const displayedRow: Ref<Labellable | null> = ref(null);
+const setDisplayedRow = (val: Labellable | null) => {
+    displayedRow.value = val;
+    // TODO: routing
+};
```
jacobtylerwalls, line 27: From speaking with @robgaston just now, FYI @chrabyrd that I think the next step will be to do what this TODO says and lift this state up somewhere so that the search component can set the displayed row (probably need to rename it displayed concept or scheme). displayed "item"? This was tricky in the controlled list manager, which had items, but we can probably use "item" here.

vitest.config.mts
```diff
@@ -16,6 +16,7 @@ function generateConfig(): Promise<UserConfigExport> {
             '**/cypress/**',
             '**/.{idea,git,cache,output,temp}/**',
             '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*',
+            'arches_lingo/constants.ts',
```
jacobtylerwalls, line 19: I'll revert this: I was misunderstanding how/when this file got covered in whole or in part.

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeBase.vue
chrabyrd, line 1: A few things here:

From my latest understanding of the mockup, the tree is a drawer not a splitter:

Screenshot 2024-10-31 at 2 40 47 PM

Now this may change, or we may decide a splitter is better, so not married to the idea but just pointing it out.

One thing the mock is showing is that the tree component is accessed by clicking a button in the subheader on many different pages. To me this says it's a lazy loaded component that should be entirely ( or near entirely ) contained, so it doesn't leak logic like `setDisplayedRow` to the components where it's called. Definitely has room for interesting prop and/or emit patterns to signal updates. This also sorta goes back on the idea that was discussed during sprint planning ( or standup? I forget ) about only showing one scheme in the tree at a time. One would imagine that if you're on the Schemes list page you'd want to be able to open the tree and see a list of all schemes.

Lastly this may just be a naming issue, but I don't love the idea of `Base` components as frontend patterns. Very similar to templates and IMO contrary to component decomposition. I visualize the concept and scheme pages similar to:

```
// pages/Concept.vue  AND pages/Scheme.vue

<div>
  <SubHeader />
  <div>
     ... stuff specific to the page
  </div>
</div>
```

with the Subheader containing the Tree and maybe a listener that tells the tree when to re-render. This is definitely an oversimplified example. Happy to chat to flesh out the idea more if need be 👍

chrabyrd (changes requested):
Overall looks good! Just a few things. Also one thing I noticed is that pulling in `main` results in some merge conflicts 😢

arches_lingo/src/arches_lingo/api.ts
```diff
@@ -79,3 +79,16 @@ export const fetchSearchResults = async (
         throw new Error((error as Error).message || response.statusText);
     }
 };
+
+export const fetchConcepts = async () => {
```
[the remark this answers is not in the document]
jacobtylerwalls: Nice. I think the original reason for defending against `response.json()` failing is because invalid routes or 404s and whatnot were responding with HTML, but we've since cleaned that up in https://github.com/archesproject/arches/pull/11167, so I'm happy to start adjusting these to be simpler ⭐

arches_lingo/src/arches_lingo/components/misc/LetterCircle.vue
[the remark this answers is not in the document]
jacobtylerwalls: Right, I mention this in the PR body. There's a choice here about whether to import it from another project, override it from another project, move it to arches-vue-utils, and honestly I've had doubts about this icon from jump, so I'd like a redesign to be on the table as well. Can we defer this to a new issue and discuss it with the group?

arches_lingo/src/arches_lingo/constants.ts
```diff
@@ -1,12 +1,32 @@
...
+import type {
+    Concept,
+    HeaderRefAndSetter,
+    UserRefAndSetter,
+} from "@/arches_lingo/types.ts";
 
 export const ANONYMOUS = "anonymous";
 export const ERROR = "error";
+export const SECONDARY = "secondary";
+export const CONTRAST = "contrast";
```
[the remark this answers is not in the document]
jacobtylerwalls: That's fine, and honestly for strings that are primevue constants, I'm not sure we need to defend against typos so aggressively, I'd be fine with just inlining them. It's the custom strings I like seeing pulled out.

arches_lingo/src/arches_lingo/components/misc/LetterCircle.vue
```diff
@@ -0,0 +1,53 @@
+<script setup lang="ts">
+import { computed } from "vue";
+
+import { dataIsScheme } from "@/arches_lingo/utils.ts";
+
+import type { Concept, Scheme } from "@/arches_lingo/types";
+
+const props = defineProps<{ labelled: Concept | Scheme }>();
+
+const color = computed(() => {
```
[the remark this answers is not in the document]
jacobtylerwalls: Yes, but as I mentioned above, I think this is a candidate for redesign, so let's not polish it?

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeBase.vue
jacobtylerwalls: > a follow-up ticket

Or better, just repurpose #111.

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeBase.vue
```diff
@@ -0,0 +1,114 @@
...
+                    systemLanguage.code,
+                ).value
+            }}
+        </h2>
+        <Button
+            :severity="shouldUseContrast() ? CONTRAST : SECONDARY"
+            :label="$gettext('Toggle hierarchy')"
+            @click="() => (showHierarchy = !showHierarchy)"
+        />
+    </div>
```
jacobtylerwalls, line 69: I'm sure this will become its own component, but I don't want to polish it until we know what we're doing. Will there be a toggle button? Does this live in a drawer?

arches_lingo/src/arches_lingo/components/misc/LetterCircle.vue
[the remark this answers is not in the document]
jacobtylerwalls: #114

jacobtylerwalls:
Thanks for the review! I think the future of a couple things need to be fleshed out a little further before working on them.

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeBase.vue
jacobtylerwalls: Removed "base" from the file name. I see the subheading becoming its own component, but I guess I'd rather not polish that until we know where the toggle button goes, otherwise that's a lot of state to pass around before firmed up.

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeSplitter.vue
chrabyrd, line 1: IMO this component should be genericized and live in the component dir, it shouldn't be a page. 

There should instead be 2 separate page components `ConceptPage.vue` and `SchemePage.vue` served at the end of each route respectively. Each should use the generic form of this. That way logic like

```
            <ConceptDetail
                v-if="displayedRow && dataIsConcept(displayedRow)"
                :concept="displayedRow as Concept"
            />
            <SchemeDetail
                v-else-if="displayedRow && dataIsScheme(displayedRow)"
                :scheme="displayedRow as Scheme"
            />
```

can be avoided.

chrabyrd (changes requested):
looks good! just a minor thing

arches_lingo/src/arches_lingo/pages/ConceptOrSchemeSplitter.vue
jacobtylerwalls: I had a worry that the splitter and tree would get destroyed on each concept click if I do that, but I'll have a look and see how hard it is.

arches_lingo/src/arches_lingo/components/tree/ConceptTree.vue
```diff
@@ -0,0 +1,287 @@
...
+        }"
+        @node-collapse="nextFilterChangeNeedsExpandAll = true"
+        @node-select="updateSelectedAndExpanded"
+    >
+        <template #default="slotProps">
+            <TreeRow
+                v-model:focused-node="focusedNode"
+                v-model:filter-value="filterValue"
+                :node="slotProps.node"
+                :focus-label="FOCUS"
```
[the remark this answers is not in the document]
jacobtylerwalls: I'll add a comment to clarify, but yeah, given that there are potentially 10,000+ concepts, I wanted to avoid running $gettext 20,000 times.

arches_lingo/src/arches_lingo/components/tree/HierarchySplitter.vue
```diff
@@ -0,0 +1,106 @@
...
+    }
+};
+// @ts-expect-error vue-tsc doesn't like arbitrary properties here
+provide(displayedRowKey, { displayedRow, setDisplayedRow });
+
+const showHierarchy = ref(true);
+</script>
+
+<template>
+    <div class="subheading">
```
jacobtylerwalls, line 53: Happy to factor this out into a component once we know whether it should also be the component that contains the Toggle Hierarchy button. (I don't think we know that yet?)

arches_lingo/src/arches_lingo/pages/ConceptPage.vue
```diff
@@ -0,0 +1,15 @@
+<script setup lang="ts">
+import { inject } from "vue";
+
+import { displayedRowKey } from "@/arches_lingo/constants.ts";
+
+import type { DisplayedRowRefAndSetter } from "@/arches_lingo/types";
+
+const { displayedRow } = inject(
+    displayedRowKey,
+) as unknown as DisplayedRowRefAndSetter;
```
chrabyrd, line 10: `as unknown as DisplayedRowRefAndSetter`?

arches_lingo/src/arches_lingo/App.vue
```diff
@@ -64,8 +74,9 @@ router.beforeEach(async (to, _from, next) => {
         <PageHeader v-if="route.meta.shouldShowNavigation" />
         <div style="display: flex; flex: auto; flex-direction: row">
             <SideNav v-if="route.meta.shouldShowNavigation" />
-            <div style="margin: 1rem; flex: auto">
-                <RouterView />
+            <div style="flex: auto">
+                <HierarchySplitter v-if="route.meta.shouldShowHierarchy" />
```
chrabyrd, line 78: Good solution! 👍

chrabyrd (commented):
not sure if this is ready for review yet, but wanted to stick my nose in anywho. Overall looks good!

jacobtylerwalls:
Oh yeah it's ready, I just push minor cleanups when I find them. The last [request](https://github.com/archesproject/arches-lingo/pull/86#event-15396641897) got collapsed out of view, haha. LMK what you think!

chrabyrd:
> Oh yeah it's ready, I just push minor cleanups when I find them. The last [request](https://github.com/archesproject/arches-lingo/pull/86#event-15396641897) got collapsed out of view, haha. LMK what you think!

oh nice ! Yeah lgtm -- just that typing question. If it ends up being a thing I'm happy to not block on it as it's a minor nit really  👍

arches_lingo/src/arches_lingo/pages/ConceptPage.vue
jacobtylerwalls: Yeah, the tooling is aggressively inferring the result of `inject()` as the first element (the ref) not the element I'm actually unpacking: the setter. Unpacking both elements didn't help either, just caused more eslint warnings.

If I find some energy after Christmas I'll try cooking up a repro for vue-tsc, maybe they'll let me know of something silly I'm overlooking!

jacobtylerwalls: See the advice to use unknown as the escape hatch:
```ts
Conversion of type 'Ref<Concept | null, Concept | null> | undefined' to type 'DisplayedRowRefAndSetter' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'Ref<Concept | null, Concept | null>' is missing the following properties from type 'DisplayedRowRefAndSetter': displayedRow, setDisplayedRowts-plugin(2352)
```

arches_lingo/src/arches_lingo/App.vue
jacobtylerwalls: I agree this is better--thanks for encouraging me down this path!

arches_lingo/src/arches_lingo/pages/ConceptPage.vue
chrabyrd: ++  --thanks for looking into it 👍

chrabyrd (approved):
LGTM! 🎸

arches_lingo/src/arches_lingo/api.ts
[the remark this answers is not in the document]
jacobtylerwalls: @chrabyrd I just ran into a case that might add some color around why I was so paranoid the first time.

Log in to lingo as tester3/Test12345!

See:
html-no-parse

This is because the `@group_required` decorator on `ConceptTreeView` (and the others) returns HTML, and we crash on calling .parse()

***
In version 8 of arches, we should probably add [custom](https://docs.djangoproject.com/en/5.1/topics/http/views/#customizing-error-views) 400, 403, 404 (maybe) and 500 handlers that sniff for whether it's an api route (prefixed with `api/` or `api.`, I guess?) and return JSON instead.

PR #91, opened 2024-09-10
Consume more tokens from arches theme, add dark mode toggle

### LINKED ISSUE #87: Arches Lingo header is <h2>
Should be `<h1>` so that we don't skip any `<h>` levels

jacobtylerwalls:
Fix header skip closes #87
Add dark mode toggle closes #91

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 -   Add login interface [#13](https://github.com/archesproject/arches-lingo/issues/13)
 -   Add front-end router [#11](https://github.com/archesproject/arches-lingo/issues/11)
--   Add dark mode toggle
+-   Add dark mode toggle [#91](https://github.com/archesproject/arches-lingo/issues/91)
 -   Add backend for search [#67](https://github.com/archesproject/arches-lingo/issues/67)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 -   Add login interface [#13](https://github.com/archesproject/arches-lingo/issues/13)
 -   Add front-end router [#11](https://github.com/archesproject/arches-lingo/issues/11)
+-   Add dark mode toggle
 -   Add backend for search [#67](https://github.com/archesproject/arches-lingo/issues/67)
 
 ### Fixed
--- a/arches_lingo/media/js/views/root.js
+++ b/arches_lingo/media/js/views/root.js
@@ -1,7 +1,5 @@
 import createVueApplication from 'arches/arches/app/media/js/utils/create-vue-application';
-import { ArchesPreset, DEFAULT_THEME } from '@/arches/themes/default.ts';
 
-import { definePreset } from '@primevue/themes';
 import { createRouter, createWebHistory } from 'vue-router';
 
 import LingoApp from '@/arches_lingo/App.vue';
@@ -12,26 +10,7 @@ const router = createRouter({
     routes,
 });
 
-const LingoPreset = definePreset(ArchesPreset, {
-    components: {
-        button: {
-            root: {
-                label: {
-                    fontWeight: 600,
-                },
-            },
-        },
-    },
-});
-
-const LingoTheme = {
-    theme: {
-        ...DEFAULT_THEME,
-        preset: LingoPreset,
-    },
-};
-
-createVueApplication(LingoApp, LingoTheme).then(vueApp => {
+createVueApplication(LingoApp).then(vueApp => {
     vueApp.use(router);
     vueApp.mount('#lingo-mounting-point');
 });
--- a/arches_lingo/media/js/views/root.js
+++ b/arches_lingo/media/js/views/root.js
@@ -1,5 +1,7 @@
 import createVueApplication from 'arches/arches/app/media/js/utils/create-vue-application';
+import { ArchesPreset, DEFAULT_THEME } from '@/arches/themes/default.ts';
 
+import { definePreset } from '@primevue/themes';
 import { createRouter, createWebHistory } from 'vue-router';
 
 import LingoApp from '@/arches_lingo/App.vue';
@@ -10,7 +12,26 @@ const router = createRouter({
     routes,
 });
 
-createVueApplication(LingoApp).then(vueApp => {
+const LingoPreset = definePreset(ArchesPreset, {
+    components: {
+        button: {
+            root: {
+                label: {
+                    fontWeight: 600,
+                },
+            },
+        },
+    },
+});
+
+const LingoTheme = {
+    theme: {
+        ...DEFAULT_THEME,
+        preset: LingoPreset,
+    },
+};
+
+createVueApplication(LingoApp, LingoTheme).then(vueApp => {
     vueApp.use(router);
     vueApp.mount('#lingo-mounting-point');
 });
--- a/arches_lingo/src/arches_lingo/App.vue
+++ b/arches_lingo/src/arches_lingo/App.vue
@@ -74,7 +74,7 @@ router.beforeEach(async (to, _from, next) => {
 
 <style scoped>
 main {
-    font-family: sans-serif;
+    font-family: system-ui, sans-serif;
     height: 100vh;
     width: 100vw;
     overflow-x: hidden;
--- a/arches_lingo/src/arches_lingo/App.vue
+++ b/arches_lingo/src/arches_lingo/App.vue
@@ -71,6 +71,8 @@ router.beforeEach(async (to, _from, next) => {
     </main>
     <Toast
         :pt="{
+            summary: { fontSize: 'medium' },
+            detail: { fontSize: 'small' },
             messageIcon: {
                 style: { marginTop: 'var(--p-toast-messageicon-margintop)' },
             },
@@ -80,7 +82,7 @@ router.beforeEach(async (to, _from, next) => {
 
 <style scoped>
 main {
-    font-family: var(--p-font-family);
+    font-family: system-ui, sans-serif;
     height: 100vh;
     width: 100vw;
     overflow-x: hidden;
--- a/arches_lingo/src/arches_lingo/App.vue
+++ b/arches_lingo/src/arches_lingo/App.vue
@@ -82,7 +82,7 @@ router.beforeEach(async (to, _from, next) => {
 
 <style scoped>
 main {
-    font-family: system-ui, sans-serif;
+    font-family: sans-serif;
     height: 100vh;
     width: 100vw;
     overflow-x: hidden;
--- a/arches_lingo/src/arches_lingo/App.vue
+++ b/arches_lingo/src/arches_lingo/App.vue
@@ -69,7 +69,13 @@ router.beforeEach(async (to, _from, next) => {
             </div>
         </div>
     </main>
-    <Toast />
+    <Toast
+        :pt="{
+            messageIcon: {
+                style: { marginTop: 'var(--p-toast-messageicon-margintop)' },
+            },
+        }"
+    />
 </template>
 
 <style scoped>
--- a/arches_lingo/src/arches_lingo/App.vue
+++ b/arches_lingo/src/arches_lingo/App.vue
@@ -80,7 +80,7 @@ router.beforeEach(async (to, _from, next) => {
 
 <style scoped>
 main {
-    font-family: system-ui, sans-serif;
+    font-family: var(--p-font-family);
     height: 100vh;
     width: 100vw;
     overflow-x: hidden;
--- a/arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
+++ b/arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
@@ -2,13 +2,12 @@
 import { ref } from "vue";
 import { useGettext } from "vue3-gettext";
 
+import { Theme } from "@primeuix/styled";
 import ToggleButton from "primevue/togglebutton";
-
-import { DEFAULT_THEME } from "@/arches/themes/default.ts";
 
 const { $gettext } = useGettext();
 
-const darkModeClass = DEFAULT_THEME.theme.options.darkModeSelector.substring(1);
+const darkModeClass = Theme.options.darkModeSelector.substring(1);
 const isDarkModeEnabled = ref(
     document.documentElement.classList.contains(darkModeClass),
 );
@@ -16,7 +15,10 @@ const isDarkModeEnabled = ref(
 function toggleDarkMode() {
     document.documentElement.classList.toggle(darkModeClass);
     isDarkModeEnabled.value = !isDarkModeEnabled.value;
-    localStorage.setItem(darkModeClass, isDarkModeEnabled.value.toString());
+    localStorage.setItem(
+        `arches.${darkModeClass}`,
+        isDarkModeEnabled.value.toString(),
+    );
 }
 </script>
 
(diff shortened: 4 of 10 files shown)


arches_lingo/src/arches_lingo/App.vue
```diff
@@ -74,7 +74,7 @@ router.beforeEach(async (to, _from, next) => {
 
 <style scoped>
 main {
-    font-family: sans-serif;
+    font-family: system-ui, sans-serif;
```
jacobtylerwalls, line 77: With this change, there's now a difference (on my local browser) between semibold and bold. Which means: we can preset the font weight for button labels to 600/semibold without it being too loud (it was looking the same as bold)

Then we'll have a default for button labels that looks good both in the controlled list manager (arches chrome) and here.

arches_lingo/src/arches_lingo/components/header/PageHeader.vue
```diff
@@ -56,8 +55,8 @@ const items = ref([
 </template>
 
 <style scoped>
-.page-header {
-    border-radius: 0;
```
jacobtylerwalls, line 60: I added this token to the core arches theme.

arches_lingo/src/arches_lingo/components/login/LoginForm.vue
```diff
@@ -65,7 +65,7 @@ const submit = async () => {
 form {
     display: flex;
     flex-direction: column;
-    gap: 1rem;
+    gap: var(--p-content-gap);
```
jacobtylerwalls, line 68: We use a lot of 1rem gaps, so I added a custom token for it in arches.

arches_lingo/src/arches_lingo/components/login/LoginLinks.vue
```diff
@@ -13,11 +15,15 @@ const { $gettext } = useGettext();
             as="a"
             :label="$gettext('Register')"
             :href="arches.urls.signup"
+            :severity="SECONDARY"
```
jacobtylerwalls, line 18: Wanted to deemphasize these buttons.

arches_lingo/src/arches_lingo/components/user/UserInteraction.vue
```diff
@@ -52,13 +52,11 @@ const greeting = computed(() => {
 </script>
 
 <template>
-    <div style="display: flex; align-items: center">
+    <div style="display: flex; align-items: center; gap: var(--p-content-gap)">
         <span v-if="user">{{ greeting }}</span>
         <Button
-            style="margin-left: 1rem"
+            :label="$gettext('Sign out')"
```
jacobtylerwalls, line 58: Without using the label prop, it wasn't using the tokens for button label font weight (not noticeable until I started overriding that weight in core arches)

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
```diff
@@ -0,0 +1,37 @@
...
+    // Remove the leading dot from the selector to get class name.
+    element!.classList.toggle(
+        DEFAULT_THEME.theme.options.darkModeSelector.substring(1),
+    );
+    isDarkModeEnabled.value = !isDarkModeEnabled.value;
+};
+</script>
+
+<template>
+    <ToggleButton
```
jacobtylerwalls, line 27: Started with a binary toggle.

I know the linked arches issue for theme switching mentions three themes. I think high contrast mode for accessibility is its own toggle (perhaps not exposed in the UI, perhaps reads from the existing system setting or the browser settings?) and will affect both light and dark modes. We'll probably want to use the primevue utilities for updating themes in that case. I don't think we want to _define new themes_ per se.

arches_lingo/src/arches_lingo/components/login/LoginForm.vue
jacobtylerwalls: But as I mentioned [here](https://github.com/archesproject/arches/pull/11319#discussion_r1754464331) it's a little debatable and we can totally revert this.

chrabyrd:
looks like some webpack failures in the CI?

jacobtylerwalls:
That's expected, since this depends on themes/default.ts in https://github.com/archesproject/arches/pull/11319. I could update this branch to depend on it, but that seems like avoidable churn?

arches_lingo/src/arches_lingo/components/header/PageHeader.vue
jacobtylerwalls: nvm, we're un-tokenizing this now 👍

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 -   Add login interface [#13](https://github.com/archesproject/arches-lingo/issues/13)
 -   Add front-end router [#11](https://github.com/archesproject/arches-lingo/issues/11)
+-   Add dark mode toggle
```
chrabyrd, line 13: may as well create/link a ticket

arches_lingo/media/js/views/root.js
```diff
@@ -10,7 +12,26 @@ const router = createRouter({
     routes,
 });
 
-createVueApplication(LingoApp).then(vueApp => {
+const LingoPreset = definePreset(ArchesPreset, {
+    components: {
+        button: {
+            root: {
+                label: {
+                    fontWeight: 600,
```
chrabyrd, line 20: I don't love the idea of overriding the font weight of all lingo buttons. Would it be a better idea to keep it standard and override on a case-by-case basis?

arches_lingo/media/js/views/root.js
```diff
@@ -10,7 +12,26 @@ const router = createRouter({
...
+                },
+            },
+        },
+    },
+});
+
+const LingoTheme = {
```
chrabyrd, line 27: If we're gonna make a theme, may as well put it in it's own file

arches_lingo/src/arches_lingo/App.vue
chrabyrd: Eh I'd say remove `system-ui`, as that varies from system to system and pulls us away from a unified experience. https://infinnie.github.io/blog/2017/systemui.html

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
```diff
@@ -41,13 +41,13 @@ const items = ref([
     display: flex;
     flex-direction: column;
     align-items: center;
-    min-width: 3rem;
+    min-width: var(--p-button-icon-only-width);
```
chrabyrd, line 44: just for my own understanding, where is this declared?

arches_lingo/src/arches_lingo/components/login/LoginLinks.vue
```diff
@@ -13,11 +15,15 @@ const { $gettext } = useGettext();
...
+            :outlined="true"
         />
         <Button
```
chrabyrd, line 21: nit, and not necessarily part of this PR, but the text in the button is underlined and arguably shouldn't be.

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
```diff
@@ -0,0 +1,39 @@
...
+import { DEFAULT_THEME } from "@/arches/themes/default.ts";
+
+const { $gettext } = useGettext();
+
+const isDarkModeEnabled = ref(
+    window.matchMedia &&
+        window.matchMedia("(prefers-color-scheme: dark)").matches,
+);
+
+function toggleDarkMode() {
```
chrabyrd, line 16: instead of classes, how do you feel with the `data-*` attr (eg data-theme)? I've seen that pretty often with other theme switchers

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
chrabyrd: yeah happy to punt 🏈

chrabyrd (changes requested):
For whatever reason, the toggle doesn't work for me. Also would it go into this ticket or a followup to have the basic search component handle theming?

arches_lingo/src/arches_lingo/components/login/LoginLinks.vue
jacobtylerwalls: My thought was that it's a better practice for links to be underlined (at least that's what I'm sure WCAG says), and this button is really a link (`as="a"`)

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
jacobtylerwalls: PrimeVue

arches_lingo/media/js/views/root.js
jacobtylerwalls: Sure, let's try it.

arches_lingo/src/arches_lingo/App.vue
jacobtylerwalls: Yeah I saw that also, but I can't remember if I saw that before or after I made this addition, 😄 . I'll back it out. You have a good point we're building a unified experience, not a blog for a law office.

arches_lingo/src/arches_lingo/components/login/LoginLinks.vue
chrabyrd: oh fair. Not to be handled in this ticket but I could see abandoning the button entirely then and just having links. But IMO wait for that order to come from someone with a stronger opinion 👍

jacobtylerwalls:
> Also would it go into this ticket or a followup to have the basic search component handle theming?

Let's do a follow up, could be good to onboard another team member onto theming stuff?

arches_lingo/src/arches_lingo/components/sidenav/SideNav.vue
jacobtylerwalls: https://primevue.org/button/#theming.tokens

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
jacobtylerwalls: I was trying to pair this with the doc'd pattern in primevue, which we already merged to core, see:

https://github.com/archesproject/arches/blob/7289bd3d5b678e6c67f6455f1c10c6df9d5659af/arches/app/src/arches/themes/default.ts#L53

jacobtylerwalls:
> For whatever reason, the toggle doesn't work for me. 

Yowza! I guess... do you have any manual overrides of the dev/8.0.x arches theme? I know on your system you did some deep dives into css reset for vue, so I'm wondering if you had any artifacts from that lying around, like css layers or something.

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
chrabyrd: oh fair

chrabyrd:
> > For whatever reason, the toggle doesn't work for me.
> 
> Yowza! I guess... do you have any manual overrides of the dev/8.0.x arches theme? I know on your system you did some deep dives into css reset for vue, so I'm wondering if you had any artifacts from that lying around, like css layers or something.

oh huh. after you pulled main into the branch things are working now 🎉

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
chrabyrd, line 1: `isDarkModeEnabled` works! (🎉 ). However, regardless of if its value, the theme when the page is loaded is always light. I was able to solve this locally via

```
onMounted(() => {
    if (isDarkModeEnabled.value) {
        toggleDarkMode();
    }
});
```
However, this may not be the right solution. There's further complexity to add that may nullify that pattern. Namely, I can set the toggle to `dark`, see the change, refresh the page, and it's automatically set back to `light`. My guess is that either cookies or localstorage need to be leveraged to persist this through page refreshes. 

If we wanted to get really fancy we could add a column to the `User` model to support preferred theme, but that's _probably_ overkill at this point. And we'd still need something on the browser to persist for anonymous user so /shrug that's not an entirely baked idea.

If this component is changed to read from cookies/localstorage on mount, it'd be more readable to change the toggle method to accept an arg so it can be explicitly set the mode. This will also support adding other themes later. However I'm not married to this pattern, YAGNI and whatnot /shrug. Regardless, reading from cookies/localstorage nullifies the simple code laid out above.

But this reveals another titchy thing: we're handling all of theming in this component. If I have a cookie ( or bit of localstorage ) saved that says I prefer dark mode, I'd expect the application to handle this regardless of the presence or absence of the header / DarkModeToggle. Like in reports and whatnot. It could be argued that the ability to switch themes should be handled in this component of course, but the ability to query a preferred theme and apply it should be handled at a deeper level. Perhaps even in core since it's touching the `<html>` element, but I don't feel strongly about that yet.

chrabyrd (changes requested):
Looks good! Some minor nits will probably fall out once this is moved out of lingo, but I'm not worrying about those now. 

There's some additional complexity that's fallen out of this review, but happy to chat or kick parts of it to a follow-up 👍

jacobtylerwalls:
> boo failing builds

I'll update the github action. Maybe we need the updateproject command to yell about this again when the github templates change?

jacobtylerwalls:
The actions are dense; better to just call out the tiny change in the upgrade instructions.

chrabyrd:
> The actions are dense; better to just call out the tiny change in the upgrade instructions.

IMO it should be automated, part of the `updateproject` command. The odds that people have changed their project's specific actions are low, but we should definitely call out a warning as part of the initial update project warning. Happy to crank this out in core. WDYT?

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
jacobtylerwalls: Good ideas all around. Updated, should retest with archesproject/arches#11667.

arches_lingo/src/arches_lingo/components/header/DarkModeToggle.vue
```diff
@@ -0,0 +1,36 @@
...
+import { ref } from "vue";
+import { useGettext } from "vue3-gettext";
+
+import ToggleButton from "primevue/togglebutton";
+
+import { DEFAULT_THEME } from "@/arches/themes/default.ts";
+
+const { $gettext } = useGettext();
+
+const darkModeClass = DEFAULT_THEME.theme.options.darkModeSelector.substring(1);
```
jacobtylerwalls, line 11: There's a little asymmetry worth deciding on.

In archesproject/arches#11667, I hard coded the `arches-dark` class name, but here I'm reading it from the default theme. The idea being that if these two ever drift (e.g. `my-arches-consultancy-dark`), it's N fewer things to update, just override base-root.htm. Thoughts?

chrabyrd: IMO we should strongly discourage ever overwriting base-root.htm. I"m not sure this pattern jives with if an application defines its own theme. However happy to punt those considerations until this toggle lives outside of Lingo

chrabyrd (commented):
Looks good! And handles persistence through refreshes.

I don't have a great answer for hard-coding on the arches core side. Perhaps there isn't one, users will just need to deal with the fact that there's a single name for dark mode toggling.

That said, I'd like to minimize hardcoding if we can get away from it. And reading `DEFAULT_THEME` from within the application doesn't cover all bases. I mucked around a bit and found something ( relatively ) graceful:


```
import { getCurrentInstance } from 'vue';
const { appContext } = getCurrentInstance();

const darkModeClass = appContext.config.globalProperties.$primevue.config.theme.options.darkModeSelector.substring(1);
```

This will at least handle for any selector for any theme fed into the application. And maybe we re-think where the dark mode class interaction is core-side once this gets moved out of core. 

I could argue that instead of living in a script tag in base-root, it could live inside `createVueApplication` so it has access to classname the theme is using and can behave appropriately. And the class still would get added before application mount.

but yeah just navel gazing I guess -- happy to punt that until later. For the time being, if you're amenable to get this change into `DarkModeToggle`, then we can merge 😄

chrabyrd (approved):
🚀 !

PR #94, opened 2024-10-01
53/54 Migrate lingo collections and graphs to use lists & reference datatype

johnatawnclementawn:


johnatawnclementawn:
@robgaston - happy to rm `lingo_lists.xlsx` in favor of `lingo_lists.json`. The excel file is there because the current iterations of `export_controlled_lists` and `import_controlled_lists` were built to handle that format.

arches_lingo/pkg/reference_data/controlled_lists/lingo_lists.xlsx
robgaston, line 1: I think we probably can just remove this file since it's not being used during load; seems like it could get confusing with the migration as well if we are loading this in package load

robgaston (changes requested):
we can just remove the xlsx file here.  otherwise, this looks good to me!

johnatawnclementawn:
For loading test data use:

# scheme
`python manage.py packages -o import_business_data -s tests/fixtures/data/aat_entries_scheme_examples.json -ow `overwrite
# concept
`python manage.py packages -o import_business_data -s tests/fixtures/data/aat_entries_concept_examples.json -ow overwrite`

PR #106, opened 2024-10-23
CI for lingo with ECS

aarongundel:
Adds an initial CI setup for lingo with ECS


arches_lingo/settings.py
```diff
@@ -54,18 +159,18 @@
...
 
 ROOT_URLCONF = "arches_lingo.urls"
 
+ELASTICSEARCH_HOSTS = [{"scheme": ES_PROTOCOL, "host": ES_HOST, "port": ES_PORT}]
 # Modify this line as needed for your project to connect to elasticsearch with a password that you generate
 ELASTICSEARCH_CONNECTION_OPTIONS = {
-    "request_timeout": 30,
-    "verify_certs": False,
-    "basic_auth": ("elastic", "E1asticSearchforArche5"),
+    "timeout": 30,
```
jacobtylerwalls, line 171: This should stay as `request_timeout` to avoid getting DeprecationWarnings about it, see archesproject/arches@a9d8e74c7f369dce2c8f4146ecffd5041c9c4d11

arches_lingo/settings.py
```diff
@@ -54,18 +159,18 @@
 UPLOADED_FILES_DIR = "uploadedfiles"
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = "--+c7*txnosqv=flep00qp+=t-xhrj%f4==r8w*n_7pm@mi%)7"
-
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+SECRET_KEY = get_optional_env_variable(
+    "ARCHES_SECRET_KEY", "--+c7*txnosqv=flep00qp+=t-xhrj%f4==r8w*n_7pm@mi%)7"
```
jacobtylerwalls, line 163: Might be nice to reuse the `django-insecure-` prefix so that anybody who tries to deploy lingo themselves and forgets to rotate this gets the built-in check, see archesproject/arches@a6d886025e4cb4cc7aa032e1a63b125c48990643, but let me know if I'm missing context

docker/settings_docker.py
```diff
@@ -0,0 +1,64 @@
...
+MODE = get_env_variable("DJANGO_MODE")
+
+DEBUG = ast.literal_eval(get_env_variable("DJANGO_DEBUG"))
+
+DATABASES = {
+    "default": {
+        "ENGINE": "django.contrib.gis.db.backends.postgis",
+        "NAME": get_env_variable("PGDBNAME"),
+        "USER": get_env_variable("PGUSERNAME"),
+        "PASSWORD": get_env_variable("PGPASSWORD"),
```
jacobtylerwalls, line 32: Can we add the options key from the main settings?

aarongundel: This file's unused.  Some of this is setting up for local development with containers - this particular file will not be returning.

arches_lingo/settings.py
aarongundel: good point.

arches_lingo/settings.py
aarongundel: thanks for this, will need to update for other projects as well.

robgaston (approved):
this looks good.  note: i had to set `DEBUG = True` in my local Django settings to run w/ development webpack build, but this is good because a default of `DEBUG = False` is probably better

PR #127, opened 2024-11-15
Update Concept and Scheme models

### LINKED ISSUE #123: Re-migrate concept->reference nodes with default values
- [x] capture default values configured for concept nodes now that reference datatype supports default values
- [x] update modeling issues to be enumerated here:
  - [x] Nodegroups with string content node and corresponding language node should use the non-localized string datatype for the content/label/statement node and a single-select reference datatype for language node:
    - [x] Scheme:
      - [x] right_statement_content + right_statement_language
      - [x] appellative_status_ascribed_name_content + appellative_status_ascribed_name_language
      - [x] statement_content_n1 + statement_language_n1
    - [x] Concept:
      - [x] statement_content + statement_language
      - [x] statement_data_assignment_statement_content + statement_data_assignment_statement_language
      - [x] appellative_status_ascribed_name_content + appellative_status_ascribed_name_language
      - [x] data_assignment_statement_content  + data_assignment_language
  - [x] Concepts should only participate in one scheme - Part of Scheme and Top Concept Of should be cardinality-1 and a single-select resource instance nodes

johnatawnclementawn:
Resolves #123 

This will break the test data - the RDM -> Lingo migration will need to be re-run, and likely updated based on new model configurations.

### DIFF
--- a/arches_lingo/etl_modules/migrate_to_lingo.py
+++ b/arches_lingo/etl_modules/migrate_to_lingo.py
@@ -354,7 +354,7 @@ class RDMMtoLingoMigrator(BaseImportModule):
                         'valid', true,
                         'value', json_build_array(json_build_object('resourceId', conceptidfrom, 'ontologyProperty', '', 'resourceXresourceId', '', 'inverseOntologyProperty', '')),
                         'source', conceptidfrom,
-                        'datatype', 'resource-instance-list'
+                        'datatype', 'resource-instance'
                     )
                 ) as value,
                 conceptidto as resourceinstanceid, -- map target concept's new resourceinstanceid to its existing conceptid
@@ -483,7 +483,7 @@ class RDMMtoLingoMigrator(BaseImportModule):
                         }
                     ],
                     "source": root_scheme,
-                    "datatype": "resource-instance-list",
+                    "datatype": "resource-instance",
                 }
             }
 
--- a/arches_lingo/etl_modules/migrate_to_lingo.py
+++ b/arches_lingo/etl_modules/migrate_to_lingo.py
@@ -293,7 +293,7 @@ class RDMMtoLingoMigrator(BaseImportModule):
         cursor.execute(
             """
             with recursive collection_hierarchy as (
-                select conceptidfrom as root_list,
+                select conceptidfrom as root_scheme,
                     conceptidto as child,
                     ARRAY[conceptidfrom] AS path,
                     0 as depth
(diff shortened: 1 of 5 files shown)


robgaston (approved):
lgtm! thanks!

johnatawnclementawn:
1. load [AFS thesaurus](https://github.com/archesproject/arches-for-science/blob/dev/2.0.x/arches_for_science/pkg/reference_data/concepts/disco_aat_thesaurus.xml)
2. start celery (I use `python -m celery -A arches_lingo worker -l info`)
3. Run RDM to Lingo ETL module for `AAT Entries` http://localhost:8000/plugins/bulk-data-manager
4. Verify number of part_of_scheme tiles created matches the number of concept resources created

arches_lingo/etl_modules/migrate_to_lingo.py
```diff
@@ -442,10 +442,22 @@ def init_relationships(
...
             resourceinstanceid = concept["concept"]
 
+            concept_has_part_of_scheme = resourceinstanceid in concepts_with_scheme
+            existing_scheme = concepts_with_scheme.get(resourceinstanceid)
+            if not concept_has_part_of_scheme:
+                concepts_with_scheme[resourceinstanceid] = root_list
+            elif concept_has_part_of_scheme and existing_scheme == root_list:
+                continue
+            elif concept_has_part_of_scheme and existing_scheme != root_list:
+                raise Exception(
```
jacobtylerwalls, line 457:
```suggestion
                raise CommandError(
```

jacobtylerwalls (approved):
+1 tested. Few things noticed along the way that can be follow-ups:

- make etl_modules an explicit package, so that it's measured by coverage
- use CommandError
- catch CommandError and surface the error message in the GUI, but may want to check with @chiatt

johnatawnclementawn:
@robgaston python coverage was previously passing because Lingo ETL Modules was not a package. Given the blocking nature of these model changes for other work in flight, I created a follow-up issue #131. Would you like to see that addressed in this PR or okay to do the follow up?

robgaston:
@johnatawnclementawn it's fine to address the coverage in the follow up ticket

arches_lingo/etl_modules/migrate_to_lingo.py
johnatawnclementawn: Updated to follow pattern in [single_csv here](https://github.com/archesproject/arches/blob/ab1e5bf7038b304cdd10f93bd1f4a27829022549/arches/app/etl_modules/import_single_csv.py#L192C1-L198C14). Will need to think about how to handle errors like this when we update this ETL module to be run via CLI, re. https://github.com/archesproject/arches/pull/10519

PR #132, opened 2024-11-20
Update backend services for nonlocalized string model changes

jacobtylerwalls:
Follow-up to #127

### Testing instructions
- Reinstall package from #127 
- Migrate to zero and back. I did this instead of creating another migration.
  - `python manage.py migrate arches_lingo zero`
  - `python manage.py migrate`
- Visit search interface, verify search works
- Verify api/concepts returns a result

### DIFF
--- a/arches_lingo/migrations/0002_load_lingo_lists.py
+++ b/arches_lingo/migrations/0002_load_lingo_lists.py
@@ -1,8 +1,9 @@
+import io
 import os
+
 from django.db import migrations
 from django.core import management
 from arches_lingo.settings import APP_ROOT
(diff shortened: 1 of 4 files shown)


arches_lingo/migrations/0002_load_lingo_lists.py
```diff
@@ -22,6 +23,7 @@ def load_lists(apps, schema_editor):
                 "controlled_lists",
                 "lingo_lists.json",
             ),
+            stdout=io.StringIO(),
```
jacobtylerwalls, line 26: This is just to silence the extra output during migrate ("n objects installed from fixture")

PR #136, opened 2024-12-03
Loosen arches requirement to >= 8.0.0a0

jacobtylerwalls:
Similar to archesproject/arches-vue-utils#4

We may want to explicitly depend on the arches model querysets work eventually, but we already depend on that transitively through arches-references. This just removes the conflicts, see failing [action](https://github.com/archesproject/arches-lingo/actions/runs/12141346874).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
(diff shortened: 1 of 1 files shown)


johnatawnclementawn (approved):
lgtm - I kicked CI off again to see if a re-run resolves an error with the 3.12 job

jacobtylerwalls:
oh i got you, never mind! ✔️

PR #152, opened 2024-12-13
adds minimum width to scheme page items

### LINKED ISSUE #145: Scheme page items should have minimum width
Screenshot 2024-12-04 at 9 54 17 AM

aarongundel:
closes #145 

Adds min-width at the splitter panel level - section level min-width is not respected at the section level.  If desired, I can delve into that, but this seemed to solve the concern in #145 faster.

### DIFF
--- a/arches_lingo/src/arches_lingo/pages/SchemePage.vue
+++ b/arches_lingo/src/arches_lingo/pages/SchemePage.vue
@@ -63,7 +63,8 @@ const getRef = (el: object | null, index: number) => {
     <Splitter>
         <SplitterPanel
             v-if="sectionVisible"
-            :size="75"
+            :size="66"
+            :min-size="33"
         >
             <template
                 v-for="(component, index) in components"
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (changes requested):
I still have the ability to make this almost as small as the original issue.
Screenshot 2024-12-13 at 5 58 25 PM

Should we bump from 25 to 33?




And can we use the prop for this instead of our own selector?
https://primevue.org/splitter/#api.splitterpanel.props.minSize

aarongundel:
Great thought, I didn't check the hierarchy since it starts as turned off.  The 25% of remaining real estate once the hierarchy is turned on is pretty small indeed.  I bumped this to 33.  I think it's still pretty ugly, but it's slightly better.  The editor will improve, I think, with forthcoming requested changes where we strike the tabbed interface.

jacobtylerwalls (approved):

PR #165, opened 2024-12-30
Implement scheme creation #157

jacobtylerwalls:
NOTE: pull the latest from the jtw/pythonic-resource-models branch

### DIFF
--- a/arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
+++ b/arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
@@ -1,22 +1,25 @@
 <script setup lang="ts">
 import { onMounted, ref } from "vue";
-import { useRoute } from "vue-router";
+import { useRoute, useRouter } from "vue-router";
 import { useGettext } from "vue3-gettext";
 import Button from "primevue/button";
 import SchemeReportSection from "@/arches_lingo/components/scheme/report/SchemeSection.vue";
 import NonLocalizedString from "@/arches_lingo/components/generic/NonLocalizedString.vue";
 import {
+    createScheme,
     fetchSchemeNamespace,
     updateSchemeNamespace,
 } from "@/arches_lingo/api.ts";
 import {
-    VIEW,
     EDIT,
-    OPEN_EDITOR,
     ERROR,
+    NEW,
+    OPEN_EDITOR,
     UPDATED,
+    VIEW,
 } from "@/arches_lingo/constants.ts";
 import { useToast } from "primevue/usetoast";
+
 import type {
     DataComponentMode,
     SchemeNamespaceUpdate,
@@ -25,8 +28,9 @@ import type {
 
 const toast = useToast();
 const { $gettext } = useGettext();
-const schemeInstance = ref<SchemeInstance>();
+const schemeInstance = ref<SchemeInstance>({});
 const route = useRoute();
+const router = useRouter();
 
 defineProps<{
     mode?: DataComponentMode;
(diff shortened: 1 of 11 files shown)


arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
```diff
@@ -35,13 +45,35 @@ onMounted(async () => {
...
+                params: { id: updated.resourceinstanceid },
+            });
+        } else {
+            updated = await updateSchemeNamespace(
+                route.params.id as string,
+                schemeInstance.value,
+            );
+        }
+        schemeInstance.value = updated;
+        emit(UPDATED);
```
jacobtylerwalls, line 63: This appears to work for now, but the update gets emitted before the router pushes the minted pk into the route, which is what the sections are using for the source of truth of the resource instance id. We might want to look into some alternatives -- e.g., if we ever implement backend logic to create additional tiles when a different tile is posted, then we could run into trouble with staleness of data in other sections, because they're not fetching data ("/new" is still in the route).

arches_lingo/src/arches_lingo/components/scheme/report/SchemeStandard.vue
```diff
@@ -62,25 +65,36 @@ async function getOptions(): Promise<ResourceInstanceReference[]> {
...
     } catch (error) {
         toast.add({
             severity: ERROR,
-            summary: $gettext("Error"),
-            detail:
-                error instanceof Error
-                    ? error.message
-                    : $gettext("Could not save the scheme standard"),
+            summary: $gettext("Error saving scheme"),
+            detail: (error as Error).message,
```
jacobtylerwalls, line 87: I can back out this change if you don't like it, I was just trying to minimize the number of translated strings ('standard' should really be a param we can feed)

arches_lingo/src/arches_lingo/components/scheme/report/SchemeStandard.vue
```diff
@@ -21,24 +22,26 @@ import {
...
+const router = useRouter();
 const selectedLanguage = inject(selectedLanguageKey) as Ref<Language>;
 const { $gettext } = useGettext();
 
 const { mode = VIEW } = defineProps<{
     mode?: DataComponentMode;
 }>();
 
-const emits = defineEmits([OPEN_EDITOR, UPDATED]);
+const emit = defineEmits([OPEN_EDITOR, UPDATED]);
```
chrabyrd, line 44: IMO should be `emits` to bring it inline with the rest of the app

arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
```diff
@@ -19,14 +28,15 @@ import type {
...
+const schemeInstance = ref<SchemeInstance>({});
 const route = useRoute();
+const router = useRouter();
 
 defineProps<{
     mode?: DataComponentMode;
 }>();
 
-const emit = defineEmits([OPEN_EDITOR, "updated"]);
+const emit = defineEmits([OPEN_EDITOR, UPDATED]);
```
chrabyrd, line 39: IMO should be `emits` to match the rest of the app ( OR the rest of the app should be `emit` to bring inline with this and the Vue docs  😄 )

chrabyrd (commented):
looks good overall 👍

arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
aarongundel: router push returns a promise.  Is awaiting that promise not an option?

arches_lingo/src/arches_lingo/components/scheme/report/SchemeStandard.vue
aarongundel: This seems wise to me.  Probably good to standardize these toasts, this is a good step toward that.

arches_lingo/src/arches_lingo/pages/SchemeList.vue
```diff
@@ -30,11 +31,11 @@ onMounted(async () => {
         });
     }
     schemes.value.unshift({
-        resourceinstanceid: "placeholder-id",
+        resourceinstanceid: NEW,
         descriptors: {
             en: {
                 name: "Create a new scheme",
```
aarongundel, line 37: Is hardcoding of this value desirable?  Seems like if new languages get added we'd have to revisit these descriptors.

chrabyrd (approved):
lgtm! 🚀

Issue #58, opened 2024-08-15
change location of `root.htm` from `templates/arches_lingo/root.htm` to `templates/views/root.htm`


chrabyrd:
Historically pathing has matched between JavaScript components that live in `media/js` and the templates they're paired with

PR #169, opened 2024-12-31
Scheme Label Editor & Reference Datatype Widget

### LINKED ISSUE #147: Scheme Labels Form/Editor
Scheme names will be set using the "Label" nodegroup:
Screenshot 2024-12-09 at 8 17 31 AM

"metatype" here should be pre-populated (I think) with a default value for semantic purposes so it does not need to be shown here.  I think that we will need to show the other fields here in this form.

Note, there are validation rules that we will want to enforce here in the API around the saving of labels.  In particular: a given language should only have one label with the type "preflabel".

Mock-up for this section:
concept_label_form_mockup

johnatawnclementawn:
Resolves #147 

Requires [Arches References #58](https://github.com/archesproject/arches-references/pull/58)


arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,141 @@
+<script setup lang="ts">
+import { defineProps, onMounted, ref } from "vue";
+import { fetchControlledListOptions } from "@/arches_lingo/api.ts";
+
+import NonLocalizedString from "@/arches_lingo/components/generic/NonLocalizedString.vue";
+import ReferenceDatatype from "@/arches_lingo/components/generic/ReferenceDatatype.vue";
+// import ResourceInstanceRelationships from "@/arches_lingo/components/generic/ResourceInstanceRelationships.vue";
```
chrabyrd, line 7: nit: remove this instead of commenting it out?

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,141 @@
...
+
+const props = withDefaults(
+    defineProps<{
+        value?: AppellativeStatus;
+    }>(),
+    {
+        value: () => ({}) as AppellativeStatus,
+    },
+);
+const appellative_status = ref(props.value);
```
chrabyrd, line 32: `camelCase` for JS consts, also why name it differently from `value`?

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,141 @@
...
+
+function onUpdate(newValue: string) {
+    console.log(newValue);
+}
+
+async function getOptions(listId: string): Promise<ControlledListItem[]> {
```
chrabyrd, line 38: nit: functions should be declared last, after all consts ( I def need to get that styleguide updated 😅 )

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,141 @@
...
+
+const languageOptions = ref<ControlledListItem[]>([]);
+const typeOptions = ref<ControlledListItem[]>([]);
+const statusOptions = ref<ControlledListItem[]>([]);
+const metatypeOptions = ref<ControlledListItem[]>([]);
+const eventTypeOptions = ref<ControlledListItem[]>([]);
+
+onMounted(async () => {
+    const [languageOpts, typeOpts, statusOpts, metatypeOpts, eventTypeOpts] =
+        await Promise.all([
```
chrabyrd, line 58: this is fine and I'm not blocking on it, but it still may hammer db depending on what the endpoint is doing. Happy to punt though

arches_lingo/src/arches_lingo/components/generic/ReferenceDatatype.vue
```diff
@@ -0,0 +1,35 @@
+<script setup lang="ts">
+import type {
```
chrabyrd, line 2: nit: types are imported last

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,87 @@
...
+        ) || item.labels.find((label) => label.valuetype_id === PREF_LABEL);
+    return preferredLabel?.value ?? "";
+};
+</script>
+
+<template>
+    <Select
+        v-model="val"
+        v-if:="!props.multiValue"
+        :show-toggle-all="!!options?.length"
```
chrabyrd, line 64: shouldn't need `!!`

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,87 @@
...
+        multiValue?: boolean;
+    }>(),
+    {
+        value: () => [],
+        options: () => [],
+    },
+);
+
+const emit = defineEmits(["update"]);
+const uriRef = ref(
```
chrabyrd, line 26: this is pretty complex, and would be more readable/maintainable if broken out into its own function.

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+            const items = props.options?.filter((item) =>
+                newVal.includes(item.uri),
+            );
+            emit("update", items ?? []);
+        }
+    },
+});
+
+const optionLabels = (item: ControlledListItem): string => {
```
chrabyrd, line 49: nit: function declarations should be `function foo() {`

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,87 @@
...
+    const preferredLabel =
```
chrabyrd, line 50: this could also be broken out a bit to improve readability

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,87 @@
...
+        :pt="{
+            emptyMessage: { style: { fontFamily: 'sans-serif' } },
+            option: { style: { fontFamily: 'sans-serif' } },
+        }"
+        :placeholder="$gettext('Select References')"
+    />
+    <MultiSelect
+        v-if="props.multiValue"
+        v-model="val"
+        :show-toggle-all="!!options?.length"
```
chrabyrd, line 77: same here

arches_lingo/src/arches_lingo/constants.ts
```diff
@@ -28,3 +28,10 @@ export const ENGLISH = {
     name: "English",
     scope: "system",
 };
+
+export const LANGUAGE_CONTROLLED_LIST = "55ce793b-a51a-4b25-811d-d08ea797f8c3";
```
chrabyrd, line 32: IMO hardcoding URIs on the frontend isn't a preferred pattern. Is there a way to get these supplied by the backend? If not or if it becomes super complex then okay, but IMO this pattern should be avoided if possible.

chrabyrd (commented):
Looks good! Just some pretty minor stuff -- happy to chat about about any of it

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
johnatawnclementawn: Two resource instance nodes in this nodegroup, just hadn't stubbed those out - they're now and import is used 👍🏼

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
jacobtylerwalls: Yeah, let's measure this before we prematurely optimize -- we can always switch to the endpoint that returns all lists and then filter on the frontend.

johnatawnclementawn: Switching to that endpoint may be helpful in avoiding hard-coded list id's because there is graph/node info... I just started working with this endpoint because I was wary of passing around the entire CLM

jacobtylerwalls: sweet!

chrabyrd (approved):
lgtm!

chrabyrd (changes requested):
whoops, was only looking at a subset of changes. One moment 😅

chrabyrd (changes requested):
code looks good but there's some functionality issues:

- the dropdowns do not work for a new scheme label

arches_lingo/src/arches_lingo/types.ts
```diff
@@ -46,13 +46,23 @@ export interface ControlledListeItemLabelValue {
 }
 
 export interface ControlledListItem {
+    item_id?: string;
     list_id: string;
     uri: string;
-    sortorder: number;
-    guide: boolean;
+    sortorder?: number;
+    guide?: boolean;
```
jacobtylerwalls, line 53: Looks like this is morphing into the referencedatatype tile shape instead of the ControlledListItem django model. If that's right, how about we remove these props altogether and rename the type?

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
```diff
@@ -0,0 +1,97 @@
...
+        ) || prefLabels[0];
+    return optionLabel?.value ?? "";
+}
+</script>
+
+<template>
+    <Select
+        v-model="val"
+        v-if:="!props.multiValue"
+        :show-toggle-all="options?.length"
```
jacobtylerwalls, line 74: Although I use `.length` in boolean contexts (maybe more than I should), this is a prop doc'd to take a boolean so let's actually pass a boolean, even if it's not happening to break right now.
```suggestion
        :show-toggle-all="options?.length > 0"
```

jacobtylerwalls: Screenshot 2025-01-08 at 2 09 18 PM

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,334 @@
...
+    ];
+    textualWorkOptions.value = await getResourceInstanceOptions(
+        fetchTextualWorkRdmSystemList,
+    );
+});
+</script>
+
+<template>
+    <!-- Label: string -->
+    <label for="">{{ $gettext("Label") }}</label>
```
jacobtylerwalls, line 191: Let's file a follow-up issue to fix the empty `for` attributes.

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,334 @@
...
+            summary: $gettext("Error saving scheme"),
+            detail: (error as Error).message,
+        });
+    }
+}
+
+async function getControlledListOptions(
+    listId: string,
+): Promise<ControlledListItem[]> {
+    const parsed = await fetchControlledListOptions(listId);
```
jacobtylerwalls, line 134: We don't have a global API client intercepting errors, so we have to manually show errors. Can be a follow-up.

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,334 @@
...
+            await router.push({
+                name: "scheme",
+                params: { id: updated.resourceinstanceid },
+            });
+        } else if (!formValue.value.tileid) {
+            await createSchemeLabel(route.params.id as string, formValue.value);
+        } else {
+            await updateSchemeLabel(
+                route.params.id as string,
+                formValue.value.tileid as string,
```
jacobtylerwalls, line 117: This was already typed well enough
```suggestion
                formValue.value.tileid,
```

jacobtylerwalls (commented):
Looking great!

arches_lingo/src/arches_lingo/components/generic/reference-datatype/ReferenceDatatypeEditor.vue
chrabyrd: would say if we're just casting to a bool then `Boolean()` is more explicit

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
jacobtylerwalls: I fixed this one

robgaston:
I'm noticing that the tables for Labels and Statements appear to have regressed on this branch, on `main` I see type and language:
Screenshot 2025-01-08 at 12 14 40 PM
but on this branch they do not appear in the table:
Screenshot 2025-01-08 at 12 13 22 PM

arches_lingo/urls.py
```diff
@@ -68,6 +71,11 @@
         SchemeLabelTileView.as_view(),
         name="api-scheme-label-tile",
     ),
+    path(
+        "api/scheme/label",
+        SchemeLabelCreateView.as_view(),
+        name="api-scheme-label-create",
```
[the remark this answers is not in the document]
johnatawnclementawn: This is the endpoint for POST'ing a single new label tile though. My impression is that you'd only be posting back one new tile at a time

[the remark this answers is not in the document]
jacobtylerwalls: It's also the endpoint for GETting the whole list, too.

> My impression is that you'd only be posting back one new tile at a time

For sure, but I'm just saying the singular makes it sound like it's a detail view for a single tile, which it isn't.

This is basically just following the example here: https://www.django-rest-framework.org/tutorial/2-requests-and-responses/

e.g.
- route: snippets/
- view: snippet_list
- GET a collection or POST a new one

arches_lingo/serializers.py
```diff
@@ -49,6 +53,35 @@ class Meta:
...
+        PREF_LABEL_LIST_ITEM = ListItem.objects.get(list_item_values__value="prefLabel")
+
+        if data:
+            new_label_language = data["appellative_status_ascribed_name_language"][0]
+            new_label_type = data["appellative_status_ascribed_relation"][0]
+
+            resource_instance_id = self.instance.resourceinstance.pk
+            resource_instance = self.instance.resourceinstance.__class__.as_model(
+                graph_slug, resource_ids=[resource_instance_id]
+            ).get(pk=resource_instance_id)
```
jacobtylerwalls, line 68: I'm looking into adding a developer convenience for this, either a method `.enrich()` and/or overwriting the `.resourceinstance` descriptor on the TileQuerySet so that it's a lazy getter that calls enrich() for you. 👀

jacobtylerwalls: (Not suggesting to block this PR on it.)

chrabyrd (approved):
Overall code looks good, some minor non-blocking nits there.

I'm noticing that I still can't add new labels, and the dropdowns powering the editor still do not show their selection if creating a new label. 

Also I'm noticing this warning in the console now:

```
runtime-core.esm-bundler.js:202 [Vue warn]: Extraneous non-props attributes (pt) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. 
  at <ReferenceDatatypeEditor value=null options= [{…}] multi-value=false  ... > 
  at <ReferenceDatatype value=null mode="edit" multi-value=false  ... > 
  at <LabelEditor value= {tileid: '7b559cf4-77a1-49b3-9116-aabffd464bd3', resourceinstance: 'b73e741b-46da-496c-8960-55cc1007bec4', nodegroup_id: 'ef87ac28-11de-11ef-9493-0a58a9feac02', sortorder: 0, provisionaledits: null, …} onUpdate=fn<update> > 
  at <SchemeLabel mode="edit" tileId="7b559cf4-77a1-49b3-9116-aabffd464bd3" onUpdated=fn<onSectionUpdate>  ... > 
  at <SchemeEditor key=0 editor-max=true editor-form="label"  ... > 
  at <SplitterPanel key=1 size=33 min-size=33  ... > 
  at <Splitter style= {height: '100%'} > 
  at <SchemePage onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< Proxy(Object) {__v_skip: true} > > 
  at <RouterView> 
  at <SplitterPanel min-size=25 style= {overflow-y: 'auto'} tabindex="-1" > 
  at <Splitter style= {height: '100%', overflow: 'hidden'} pt= {gutter: {…}} > 
   at <HierarchySplitter key=0 > 
  at <App>
```

but if these are planned to be addressed in followup tickets, I won't block.

arches_lingo/serializers.py
johnatawnclementawn: @jacobtylerwalls I swear this was working yesterday, but when I try to create a new label for an existing scheme self.instance is `None`. Would you be able to take a look at that in my absence?

arches_lingo/src/arches_lingo/components/generic/non-localized-string/NonLocalizedStringEditor.vue
```diff
@@ -2,7 +2,10 @@
 import { ref, watch } from "vue";
 import InputText from "primevue/inputtext";
 
-const props = defineProps<{ value: string }>();
+const props = defineProps<{
+    value: string;
+    ptId?: string;
```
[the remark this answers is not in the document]
johnatawnclementawn: Primevue `InputText` doesn't take an ariaLabeledBy prop like some of their other components, so instead I gave that component an Id, hence pass-through Id (ptId). Fixed in latest, but if we start to use these pass-throughs a lot more, I think pt as an object (pattern that primevue uses) is better than this.

arches_lingo/serializers.py
jacobtylerwalls: Done! I think you and I were only testing updates, not creates. I also added a little developer convenience for now.

arches_lingo/serializers.py
```diff
@@ -49,6 +53,35 @@ class Meta:
...
+            current_labels = resource_instance.appellative_status
+
+            for label in current_labels:
+                label_language = label.appellative_status_ascribed_name_language[0]
+                label_type = label.appellative_status_ascribed_relation[0]
```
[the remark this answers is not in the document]
jacobtylerwalls: Added a fix for this, but we probably want a dataclass or something to prevent this awkward double fallback.

arches_lingo/serializers.py
```diff
@@ -49,6 +53,35 @@ class Meta:
...
+                if (
+                    data["tileid"] != label.tileid
+                    and new_label_type["uri"] == PREF_LABEL_LIST_ITEM.uri
+                    and label_type["uri"] == PREF_LABEL_LIST_ITEM.uri
+                    and label_language["uri"] == new_label_language["uri"]
+                ):
+                    raise ValidationError(
+                        "A preferred label with the same language already exists for this scheme."
```
jacobtylerwalls: I localized this and borrowed the wording from arches-references to prevent creating a second translatable string.

arches_lingo/src/arches_lingo/components/generic/LabelEditor.vue
```diff
@@ -0,0 +1,390 @@
...
+} from "@/arches_lingo/types.ts";
+import type { Language } from "@/arches_vue_utils/types.ts";
+
+const emit = defineEmits(["update"]);
+const toast = useToast();
+const route = useRoute();
+const router = useRouter();
+const selectedLanguage = inject(selectedLanguageKey) as Ref<Language>;
+const { $gettext } = useGettext();
+const formId = (route.params.id as string) ?? Math.random().toString();
```
jacobtylerwalls, line 49: TIL that Vue 3.5 ships a `useId()` hook for this: https://vuejs.org/api/composition-api-helpers.html#useid

That's likely better than rolling our own.

Also: why wouldn't we put this label inside the `<DateDatatypeEditor>` and perhaps use `<label for=` like you had before, instead of passing this down through all these layers?

jacobtylerwalls: Pushed an update to use this

arches_lingo/serializers.py
```diff
@@ -49,6 +53,35 @@ class Meta:
         root_node = "appellative_status"
         fields = "__all__"
 
+    def validate(self, data):
+        data = super().validate(data)
+        graph_slug = self.Meta.graph_slug
+        PREF_LABEL_LIST_ITEM = ListItem.objects.get(list_item_values__value="prefLabel")
```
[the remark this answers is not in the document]
jacobtylerwalls: Pushed an update for this, a bit silly, but there are probably better things we can do in the future around identifying a list with a natural key (should name be unique, assuming it's not localized?)

jacobtylerwalls:
> I'm noticing that I still can't add new labels, and the dropdowns powering the editor still do not show their selection if creating a new label.

I fixed this by reverting the revert @johnatawnclementawn added after the demo. Johnathan, let me know if there's a reason you wanted to revert and let's discuss 👍 

> Also I'm noticing this warning in the console now

Fixed in [35a1237](https://github.com/archesproject/arches-lingo/pull/169/commits/35a1237141907ba5161174d9508abf81ddebb82d)

Issue #172, opened 2025-01-03
eslint warnings don't fail the pre-commit hook or CI


jacobtylerwalls:
eslint warnings (vs. errors) don't fail the build or CI.  I think it would be better to either silence messages we don't care about or elevate them to errors, otherwise they'll just accumulate.

Our eslint config pulls in some rules configured at the warning level, here:
https://github.com/archesproject/arches-lingo/blob/e8f9a001256a508b9240e4abb5fd79346b478e2a/eslint.config.mjs#L11

If you elevate one of them to an error, you get a bunch of fixes generated by the pre-commit hook. (I've also tinkered with this particular option following this [instruction](https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#vuehtml-self-closing) for prettier compat.)

```diff
diff --git a/eslint.config.mjs b/eslint.config.mjs
index ab51240..130a4ee 100644
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -38,6 +38,9 @@ export default [
         },
         "rules": {
             "semi": ["error", "always"],
+            "vue/html-self-closing": [
+                "error", { "html": { "void": "any" } }
+            ],
         },
     },
 ];
```

I'll PR the fixes so you can see it.

***
But we already have some other warnings in the build if you `npm run eslint:check`:

```js
/Users/.../prj/arches-lingo/arches_lingo/src/arches_lingo/components/generic/resource-instance-relationships/ResourceInstanceRelationshipsEditor.vue
  11:9  warning  Prop 'val' requires default value to be set  vue/require-default-prop

✖ 1 problem (0 errors, 1 warning)
```

☝️ seems like something would have been nice to address at the source

/cc @chrabyrd

PR #173, opened 2025-01-03
Show effect of vue/html-self-closing rule re #172

jacobtylerwalls:
Refs #172, just to spur discussion

see this particular "strongly recommended" rule [here](https://eslint.vuejs.org/rules/html-self-closing.html), but the point is that there are other rules at the warning level flying under radar

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -38,6 +38,9 @@ export default [
         },
         "rules": {
             "semi": ["error", "always"],
+            "vue/html-self-closing": [
+                "error", { "html": { "void": "any" } }
+            ],
         },
     },
 ];
--- a/arches_lingo/src/arches_lingo/components/generic/ControlledListItem.vue
+++ b/arches_lingo/src/arches_lingo/components/generic/ControlledListItem.vue
@@ -17,6 +17,6 @@ defineEmits(["update"]);
         <div v-if="mode === VIEW">
             <ControlledListItemViewer :value="value" />
         </div>
-        <div v-if="mode === EDIT"></div>
+        <div v-if="mode === EDIT" />
     </div>
 </template>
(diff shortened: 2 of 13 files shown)


eslint.config.mjs
```diff
@@ -38,6 +38,9 @@ export default [
         },
         "rules": {
             "semi": ["error", "always"],
+            "vue/html-self-closing": [
+                "error", { "html": { "void": "any" } }
```
jacobtylerwalls, line 42: If we don't like the `<div>` elements being touched, there should be a way to turn this off with `normal: any`

chrabyrd:
Yeah I'd say this deserves a larger discussion, as it deviates from the standard ruleset. If the change is agreed upon then it should also happen in core and whatnot.

On this I guess I tend to lean in the direction that it doesn't need programmatic enforcement -- nothing's breaking right?

jacobtylerwalls:
For sure we'd do it in core, this is just opened against lingo where we have Vue code to look at.

I guess I wouldn't say it deviates from the standard ruleset -- we have `eslint-config-prettier` turning _off_ some recommended rules just for the sake of prettier compat, but they explain how to get it working again [here](https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#vuehtml-self-closing).

And then the other rules are not shown in the diff because they're not auto-fixable, but since they're emitting warnings they are part of the standard ruleset and I think we should decide whether to turn 'em up to errors. Otherwise they just become things everybody ignores. Thoughts?

jacobtylerwalls:
> On this I guess I tend to lean in the direction that it doesn't need programmatic enforcement -- nothing's breaking right?

That's totally fair. For another view:
https://dev.to/thawkin3/eslint-warnings-are-an-anti-pattern-33np

PR #176, opened 2025-01-07
removes trusted authorities, fixes text for card n adds

aarongundel:
mostly cleanup/polish


arches_lingo/src/arches_lingo/components/scheme/report/SchemeNamespace.vue
```diff
@@ -110,6 +114,11 @@ function onNamespaceNameUpdate(val: string) {
         <div v-if="!mode || mode === VIEW">
             <SchemeReportSection
                 :title-text="$gettext('Scheme Namespace')"
+                :button-text="
```
chrabyrd, line 117: not sure about this. Since it's cardinality 1 should the button be hidden instead? Since I'm assuming the table row with have an edit interaction like cardinality-n tables

chrabyrd (commented):
looks good, just a `nit` and a question

chrabyrd (approved):
everything looks good, won't block on minor nits 👍

Issue #177, opened 2025-01-07
 scheme notes editor


aarongundel:
create a scheme notes editor, similar to label editor.

PR #181, opened 2025-01-09
Adds note editor (#177)

aarongundel:


arches_lingo/src/arches_lingo/components/generic/NoteEditor.vue
```diff
@@ -0,0 +1,243 @@
...
+    statusOptions.value = statusOpts;
+    metatypeOptions.value = metatypeOpts;
+    eventTypeOptions.value = eventTypeOpts;
+    groupAndPersonOptions.value = await getResourceInstanceOptions(
+        fetchGroupRdmSystemList,
+    );
+    groupAndPersonOptions.value = [
+        ...(groupAndPersonOptions.value || []),
+        ...(await getResourceInstanceOptions(fetchPersonRdmSystemList)),
+    ];
```
chrabyrd, line 152: this could stand to be refactored

chrabyrd (commented):
overall looks good just 1 small thing

arches_lingo/src/arches_lingo/components/generic/NoteEditor.vue
aarongundel: yanked out select initialize into its own method.

arches_lingo/templates/arches_urls.htm
```diff
@@ -16,6 +16,7 @@
     api_scheme_label_create="{% url 'api-scheme-label-create' %}"
     api_scheme_note='(resourceid)=>{return "{% url "api-scheme-note" "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa" %}".replace("aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", resourceid)}'
     api_scheme_note_tile='(resourceid, tileid)=>{return "{% url "api-scheme-note-tile" "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa" "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaab" %}".replace("aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", resourceid).replace("aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaab", tileid)}'
+    api_scheme_note_create="{% url 'api-scheme-note-create' %}"
```
chrabyrd, line 19: should this just fall under the api_scheme_note as a different verb?

chrabyrd (approved):
looks good! One minor question that doesn't really need to be dealt with here

arches_lingo/templates/arches_urls.htm
aarongundel: It looks like it, but that doesn't actually work.  The difference is in pythonic_models.py - you'll see there are different views used for these routes.  The ListCreateView is necessary here because we're dealing with a list of tiles.  If I was to try a patch or a put against api_scheme_note (which uses RetrieveUpdateDestroyAPIView) I'd blow away the rest of the list.

arches_lingo/src/arches_lingo/components/generic/NoteEditor.vue
```diff
@@ -0,0 +1,372 @@
...
+const valueRef = toRef(props, "value");
+
+const formValue = computed({
+    get: () => valueRef.value,
+    set: (newVal: SchemeStatement) => {
+        valueRef.value = newVal;
+    },
+});
+
+onMounted(async () => {
```
chrabyrd, line 74: style nit: `onMounted(initializeSelectOptions)`

chrabyrd (approved):
looks good! One small style nit ( and some as-of-yet unenforceable styleguide stuff ) but nothing worth blocking on 🚀

PR #197, opened 2025-01-29
Fix id attribute on `<p>` elements

jacobtylerwalls:
Follow-up to 7cec0ec.

Some of these got botched in a find/replace mishap.


johnatawnclementawn (approved):
Made small change to non-localized string because primevue's implementation doesn't support aria-labeled by

Passes accessibility wave audit 👍🏼

PR #208, opened 2025-02-05
Update graphs for correct tile keys, cope with reference datatype changes

jacobtylerwalls:
This is the partner PR to archesproject/arches-references#59.

 - Fix `listid` -> `list_id` in models (default value for references dt nodes)
 - Use the dataclass syntax (`.uri` not `["uri"]` assuming archesproject/arches-references#59 merges)
 - Rewrite the `validate_appellative_status()` method for explicitness, fewer nested fallbacks like `next(iter(...`

Should wait to merge this until archesproject/arches-references#59.

### DIFF
--- a/arches_lingo/serializers.py
+++ b/arches_lingo/serializers.py
@@ -4,6 +4,7 @@ from rest_framework.exceptions import ValidationError
 from arches.app.models.models import ResourceInstance, TileModel
 from arches.app.models.serializers import ArchesModelSerializer, ArchesTileSerializer
 
+from arches_references.datatypes.datatypes import ReferenceDataType
 from arches_references.models import ListItem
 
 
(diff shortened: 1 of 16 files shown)


arches_lingo/serializers.py
```diff
@@ -63,30 +64,37 @@ def validate_appellative_status(self, data):
             raise ValidationError(msg)
 
         if data:
```
chrabyrd, line 66: not blocking on this, but this function looks like it could be broken out into other functions?

johnatawnclementawn (changes requested):
I think there are a few changes for the rdm system models and corresponding example data that were missed in the find+replace for listid -> list_id.

johnatawnclementawn (approved):

PR #211, opened 2025-02-07
Remove parent tile workaround

jacobtylerwalls:
This branch removes the workarounds for parent tile linking and the last custom serializers.
***
Example of nested tile creation with a cardinality 1 parent:

```json
{
    "rights": {
        "right_statement": {
             "right_statement_content": "foo"
        }
    }
}
```

And then with a cardinality-N parent, you can test the following example, but would need to use Concept graph rather than Scheme graph
(notice that the browsable API takes a while to respond because we turned off default pagination, which we might want to revisit, so add it back it globally if you wish by adding the following to settings.py and removing the `pagination_class = None` in  he view)
```py
REST_FRAMEWORK = {
    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
    'PAGE_SIZE': 100  # or settings.MAX_API_PAGE_SIZE
}
```
POST to http://127.0.0.1:8000/api/lingo/concept
```json
{
    "statement": [
        {
            "statement_content": "Foo Statement",
            "statement_data_assignment_statement": {
                "statement_data_assignment_statement_content": "foo"
            }
        },
        {
            "statement_content": "Bar Statement",
            "statement_data_assignment_statement": {
                "statement_data_assignment_statement_content": "bar"
            }
        }
    ]
}
```

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import vue from "@vitejs/plugin-vue";
 
+import { fileURLToPath } from 'url';
 import { defineConfig } from 'vite';
 
 import type { UserConfigExport } from 'vite';
@@ -9,7 +10,10 @@ import type { UserConfigExport } from 'vite';
 
 function generateConfig(): Promise<UserConfigExport> {
     return new Promise((resolve, reject) => {
+        const filePath = path.dirname(fileURLToPath(import.meta.url));
+
         const exclude = [
+            '**/*.d.ts',
             '**/node_modules/**',
             '**/dist/**',
             '**/install/**',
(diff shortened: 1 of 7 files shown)


vitest.config.mts
chrabyrd, line 1: If you're making changes to these files, could you do it on a separate core branch? There's method behind the madness and i want to be sure everything still works ( eg new projects with no frontend tests should still score 100% test coverage )

chrabyrd (changes requested):
looks good, just need to make sure this doesn't destabilize testing

vitest.config.mts
chrabyrd: whoops this is just copied from core. Approved!

chrabyrd (approved):
lgtm 👍

PR #219, opened 2025-02-20
Cherry-pick fix for missing python names, update import

jacobtylerwalls:
This is fixed on other branches but don't want main to be unstable while waiting for those reviews.


jacobtylerwalls:
Also: re stability of main, fixed an import to use the new name on the core branch

aarongundel (approved):
works for me!

PR #220, opened 2025-02-20
Remove tileId param from upsertLingoTile #218

### LINKED ISSUE #218: upsertLingoTile isn't very clear about where the tile ids should be provided
`upsertLingoTile` takes a tileId param but also tileIds in the json payload, which will confuse users. Have a look at improving this and check the other api.ts functions, too.

jacobtylerwalls:
Closes #218 

Having a param to provide the tileid was misleading, when the format we're exchanging in both directions expects tileid to be a prop on the tile data object.

### DIFF
--- a/arches_lingo/src/arches_lingo/api.ts
+++ b/arches_lingo/src/arches_lingo/api.ts
@@ -1,7 +1,7 @@
 import arches from "arches";
 import Cookies from "js-cookie";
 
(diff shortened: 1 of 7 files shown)


chrabyrd:
👋 ! Mind resolving the new merge conflicts?

chrabyrd (approved):
lgtm! 🚀

PR #213, opened 2025-02-12
Cope with serializer format change

jacobtylerwalls:
Partner to https://github.com/archesproject/arches-references/pull/63

Existing functionality seems to work okay, but you can also now post back an array of list item ids as we discussed.

PR #221, opened 2025-02-20
Reimplement tile validation logic as function #215

### LINKED ISSUE #215: Further genericize serializers
The LabelEditor Form talks to the `LingoTileSerializer`. However there is custom logic in the serializer and if we're going to create a generic Card/Form component, we'd need a more generic serializer.

jacobtylerwalls:
Closes #215

Removes last custom serializers by moving validation logic to an arches function.

In future tranches of work, we could consider solutions for making function creation less verbose. Even FunctionXGraph could be reconsidered.

Apply the migration, test validation check still occurs:
- [ ] test lingo app
- [ ] test resource editor, creating new tile
- [ ] test resource editor, updating tile


arches_lingo/serializers.py
```diff
@@ -1,57 +0,0 @@
...
-
-from arches.app.models.serializers import ArchesTileSerializer
-
-from arches_controlled_lists.datatypes.datatypes import ReferenceDataType
-from arches_controlled_lists.models import ListItem
-
-
-class LingoTileSerializer(ArchesTileSerializer):
-
-    def validate_appellative_status(self, data):
```
jacobtylerwalls, line 12: An arches app developer will have a choice, either implement logic in their new APIs only, like we were doing before, or write an arches function that adds compatibility with the legacy resource editor.

arches_lingo/functions/validate_appellative_status.py
```diff
@@ -0,0 +1,56 @@
...
+from arches_controlled_lists.datatypes.datatypes import ReferenceDataType
+from arches_controlled_lists.models import ListItem
+
+
+class ValidateAppellativeStatus(BaseFunction):
+    def save(self, tile, request=None, context=None):
+        if not hasattr(tile, "appellative_status_ascribed_name_language"):
+            tile._enrich(graph_slug="scheme", only=["appellative_status"])
+        if (
+            new_label_languages := ReferenceDataType().to_python(
```
jacobtylerwalls, line 15: This little dance with the to_python() stuff might change pending #213.

johnatawnclementawn (commented):
I think this is slightly problematic in the context of the bulk data manager. IIRC, Arches functions are not invoked during any(?) executions of ETL modules in the BDM, and the primary path for getting RDM data into Lingo is the bulk data manager. Perhaps we could execute the function calculation as a step in `migrate_to_lingo`, but that would likely happen after the tiles are already created... how do we fail gracefully if there are multiple prefLabels for a concept in the same language?

robgaston:
@johnatawnclementawn thanks for bringing this up.  in theory, the RDM will have already enforced this particular validation prior to loading to lingo, so it's not such a concern in that particular case, but we would like to find a solution that works in all of these places.  @chiatt would it be possible to implement these validation-style pre-save functions (perhaps selectively) on load of data via the bulk data manager?

chiatt:
> @johnatawnclementawn thanks for bringing this up. in theory, the RDM will have already enforced this particular validation prior to loading to lingo, so it's not such a concern in that particular case, but we would like to find a solution that works in all of these places. @chiatt would it be possible to implement these validation-style pre-save functions (perhaps selectively) on load of data via the bulk data manager?

I think @johnatawnclementawn could probably execute the function as part of the staging process in his module. The results of the function could be saved to the staging record with indication of an error if there is any.

jacobtylerwalls:
Tests are failing because the migration adds a function but the graph hasn't been loaded yet.

Is this the time to start looking into expressing graph loads as migration operations?

robgaston:
I think it would be nice to define the entire "package load" as the initial (?) migration for Lingo. This would allow us to order these operations accordingly

arches_lingo/functions/validate_appellative_status.py
```diff
@@ -0,0 +1,56 @@
...
+from arches.app.functions.base import BaseFunction
+
+from arches_controlled_lists.datatypes.datatypes import ReferenceDataType
+from arches_controlled_lists.models import ListItem
+
+
+class ValidateAppellativeStatus(BaseFunction):
+    def save(self, tile, request=None, context=None):
+        if not hasattr(tile, "appellative_status_ascribed_name_language"):
+            tile._enrich(graph_slug="scheme", only=["appellative_status"])
```
jacobtylerwalls, line 13: I just deleted this function, so let me look into whether it's truly needed...

chrabyrd:
closing this as unplanned for now

PR #240, opened 2025-02-28
Depend on arches-querysets

jacobtylerwalls:
This branch depends on `arches-querysets` for the REST serializers and related functionality from the core branch we've been using.

This branch sets the arches dependency back to dev/8.0.x.


README.md
```diff
@@ -41,6 +41,11 @@
     pip install -e ../arches
     ```
 
+1. Create a settings_local.py file with `DEBUG=True`:
```
johnatawnclementawn, line 44: Would it be worth it to specify that this should be run in `arches-lingo/arches_lingo`?

johnatawnclementawn (approved):

jacobtylerwalls:
Action is failing since arches-controlled-lists@rough-move-reference-select-widget hasn't yet removed its pin to arches@jtw/pythonic-resource-models, but it will once that merge conflict is resolved 👍

PR #250, opened 2025-03-05
reverts concept statement null typing

chrabyrd:
This change was causing ts errors. reverting!

### DIFF
--- a/arches_lingo/src/arches_lingo/types.ts
+++ b/arches_lingo/src/arches_lingo/types.ts
@@ -106,8 +106,8 @@ export interface ConceptStatement extends TileData {
     statement_data_assignment_object_used?: ResourceInstanceReference[];
     statement_data_assignment_actor?: ResourceInstanceReference[];
     statement_data_assignment_type?: ControlledListItem[];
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
The serializer returns `null` for absent node values, since that's what's in the tile. Shouldn't we adjust the typing of the widget values instead of writing a type that doesn't match the API response? (Then the `undefined` is really just there to support omitting k/v pairs when posting back, but if that's not something the linter cares about, we can just maybe drop undefined?)

chrabyrd:
> The serializer returns `null` for absent node values, since that's what's in the tile. Shouldn't we adjust the typing of the widget values instead of writing a type that doesn't match the API response? (Then the `undefined` is really just there to support omitting k/v pairs when posting back, but if that's not something the linter cares about, we can just maybe drop undefined?)

no that's fair, on further thought this should be at the widget level for sure. We're appropriately nullguarding in lingo but there'll be no guarantee that once these become widely used that will always be the case. Closing this in favor of a widget-based solution && accessory lingo ticket to bring `ConceptStatement` and `SchemeStatement` in line with each other 👍

PR #255, opened 2025-03-07
Add toast for fetching errors, #235

njkim:
Handle the fetching errors with toast notification, #235

### DIFF
--- a/arches_lingo/src/arches_lingo/components/concept/ConceptLabel/ConceptLabel.vue
+++ b/arches_lingo/src/arches_lingo/components/concept/ConceptLabel/ConceptLabel.vue
@@ -39,9 +39,8 @@ onMounted(async () => {
     ) {
         const sectionValue = await getSectionValue();
         tileData.value = sectionValue.aliased_data[props.nodegroupAlias];
-    } else {
-        isLoading.value = false;
-    }
+     }
+    isLoading.value = false;
 });
 
 async function getSectionValue() {
@@ -53,8 +52,6 @@ async function getSectionValue() {
         );
     } catch (error) {
         fetchError.value = error;
-    } finally {
-        isLoading.value = false;
     }
 }
 </script>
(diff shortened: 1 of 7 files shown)


chrabyrd (changes requested):
This looks good but I don't think we should `toast` for section fetch failures like this, as there's nothing telling the user which section failed. I think it'd be better to show the user a `Message` component instead

arches_lingo/src/arches_lingo/components/scheme/SchemeStandard/SchemeStandard.vue
```diff
@@ -38,8 +40,6 @@ onMounted(async () => {
         const sectionValue = await getSectionValue();
         tileData.value = sectionValue[props.nodegroupAlias];
     }
-
-    isLoading.value = false;
```
chrabyrd, line 42: because `getSectionValue` is within an if-statement, removing `isLoading.value = false` from here can result in the component never leaving the loading state.

chrabyrd (changes requested):

chrabyrd (changes requested):
overall looks good just 1 minor thing

arches_lingo/src/arches_lingo/components/concept/ConceptLabel/ConceptLabel.vue
```diff
@@ -82,5 +83,11 @@ async function getSectionValue() {
             :resource-instance-id="props.resourceInstanceId"
             :tile-id="props.tileId"
         />
+        <Message
+            v-if="fetchError"
+            severity="error"
+            size="small"
+            >{{ fetchError.message }}
+        </Message>
```
chrabyrd, line 91: I believe this is in the wrong place. it would still show the viewer/editor in addition to the error message. instead it should be:

```
<ProgressSpinner v-if>
<Message v-else-if>
<template v-else>
```

also nit: the message format is a little wonky:

```
        <Message
            v-if="fetchError"
            severity="error"
            size="small"
        >
            {{ fetchError.message }}
        </Message>
```
```

chrabyrd (changes requested):

chrabyrd (approved):
lgtm!

PR #257, opened 2025-03-10
attempt to fix action

aarongundel:


jacobtylerwalls (approved):
Not doing this in the project template, b/c the source of this issue is the way we're defining DEBUG in this project:
```py
DEBUG = bool(get_optional_env_variable("ARCHES_DJANGO_DEBUG", False))
```

If that makes its way into the template, then we may want to revisit this, e.g. adding `--skip-checks` here instead, but that might need more thought.

+1 to merge this & unblock, thanks for the quick pickup!

PR #262, opened 2025-03-21
Create tile sort order during RDM to Lingo migration

### LINKED ISSUE #260: Create tile sortorder during RDM to Lingo migration
Tile sort order is [now captured in the bulk-data-manager](https://github.com/archesproject/arches/pull/11638). In the RDM to Lingo migration module, we should create tile sortorder so that Schemes and Concepts with multiple labels, identifiers, and notes are sorted explicitly.

johnatawnclementawn:
Resolves #260


robgaston (approved):
This looks good to me - thanks!

PR #263, opened 2025-03-24
Updates for digital object for localized string

aarongundel:
change statement content and name content to non localized string


johnatawnclementawn (approved):
LGTM 👍🏼  - there are a few remaining `string` nodes in a handful of Lingo graphs, but let's get this merged in to unblock you and I can take care of those in a separate PR

PR #283, opened 2025-04-12
Fix concept label save

jacobtylerwalls:
In anticipation of merging archesproject/arches-controlled-lists#63.

Similar changes already merged for scheme label save in #213. I was under the impression that we were planning to merge all these branches early in the week to minimize disruption, but what I appear to have done instead is cause disruption. Apologies.

Concepts should now save labels just like schemes. Test with archesproject/arches-controlled-lists#63.

### DIFF
--- a/arches_lingo/src/arches_lingo/components/concept/ConceptLabel/components/ConceptLabelEditor.vue
+++ b/arches_lingo/src/arches_lingo/components/concept/ConceptLabel/components/ConceptLabelEditor.vue
@@ -61,9 +61,7 @@ async function save(e: FormSubmitEvent) {
     isSaving.value = true;
 
     try {
-        const formData = Object.fromEntries(
-            Object.entries(e.states).map(([key, state]) => [key, state.value]),
-        );
+        const formData = e.values;
 
         let updatedTileId;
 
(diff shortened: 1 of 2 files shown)


chrabyrd (changes requested):
Nice!

This brings it inline with the Scheme editor. However, in both cases when initially creating the resource, none of the values are written to the tiles. With Schemes, it means to first create the resource, then re-edit it to add the info.

With concepts, this bricks the system. Are you able to confirm the behavior?

jacobtylerwalls:
Taking a look

jacobtylerwalls:
Sourced it to the top of `_update_tile_for_grouping_node()` in arches-querysets, which makes sense, as the symptom you reported was that the node values weren't being written to the tiles. I'll use this opportunity to add a regression test and let you know when it's ready for a look 👍

jacobtylerwalls:
Ah no, that was just a symptom! The issue was wayyy back in the aliased_data deserialization like I mused yesterday.

I pushed a fix to archesproject/arches-querysets#12. Scheme and concept creation work for me now when I use this branch of lingo (or merge it into your concept creation branch).

There is one more bug I need to look at in archesproject/arches-querysets#10 that you might notice, but it's a little different.

chrabyrd (approved):
lgtm!

PR #285, opened 2025-04-15
Update URIs in default values

johnatawnclementawn:
In #262, we updated the Lingo lists in the package with their updated versions from George and Phil. However, the default values for reference nodes were not updated, this does that.


johnatawnclementawn:
@robgaston this is something I noticed a few days ago. It's pretty inconsequential, but if you have a quick second to bless it, would be nice.

robgaston (approved):
lgtm - thanks!

PR #299, opened 2025-05-01
rough out

chrabyrd:


arches_lingo/database_routers.py
```diff
@@ -0,0 +1,45 @@
...
+            write_database_alias = alias
+            break
+
+    read_database_aliases = [
+        alias
+        for alias, _database in settings.DATABASES.items()
+        if _database.get("READ_MODE") is True
+    ]
+
+    def db_for_read(self, model, **hints):
```
aarongundel, line 23: We should probably prefer the database hinted at if it's in `hints`.  I think this is what's making these databases sticky.  And we want them to be sticky to work around replication lag.

aarongundel: See https://github.com/django/django/blob/main/django/db/utils.py

chrabyrd:
closing as unplanned

PR #301, opened 2025-05-01
Update Lingo package and example data

johnatawnclementawn:
This PR does three things:
1. Fixes a bug with the RDM to Lingo migration that did not filter skos:related relations
2. Purges old thesauri xml files from the Lingo package
    - All reference data now live in pkg/reference_data/controlled_lists & all Lingo graphs are no longer using concept/concept-list datatype 
4. Replaces old test data with FISH Thesauri that have been migrated to Lingo resources and packaged in django fixtures.

To load the new fixtures, first reload the Lingo package, then use this command:
```
python manage.py loaddata tests/fixtures/data/FISH_Thesauri_example_data_resources.json tests/fixtures/data/FISH_Thesauri_example_data_tiles.json
```
Loading should take <2 minutes


robgaston (approved):
lgtm

PR #303, opened 2025-05-06
Extract all hierarchy paths

jacobtylerwalls:


njkim (commented):
This looks good. It works as expected when the concept of interest has two parents.
However, if the concept has one parent and also a top concept of that scheme, the response only seem to include the first lineage.
The second lineage (the top concept of the scheme with no parent concept) does not seem to be included in the response.

njkim (approved):
This looks good. It works as expected when the concept of interest has two parents.
However, if the concept has one parent and also a top concept of that scheme, the response only seem to include the first lineage.
The second lineage (the top concept of the scheme with no parent concept) does not seem to be included in the response.

njkim:
Works great.

PR #305, opened 2025-05-07
Remove VwLabelValue

jacobtylerwalls:
This view predated arches-querysets and was the result of experimenting with materialized views. Once we unmaterialized it, It reduces to just a single select statement, which we don't really need a view for.

Changed to use SemanticTile from arches-querysets


tests/tests.py
```diff
@@ -35,85 +33,42 @@
...
 
+# these tests can be run from the command line via
+# python manage.py test tests.tests --settings="tests.test_settings"
+
 
 class ViewTests(TestCase):
-    @classmethod
-    def mock_concept_and_scheme_graphs(cls):
-        """Bootstrap just a few nodes as an alternative to loading the entire package."""
+    graph_fixtures = ["scheme.json", "concept.json"]
```
jacobtylerwalls, line 41: We finally outgrew this. Too bad the package takes a while to load, but such is life 🍃

njkim (approved):
lgtm

PR #348, opened 2025-06-18
Fixes missing cardxnodexwidget records for rdm system models

### LINKED ISSUE #293: `digital_object_rdm_system` does not have `card_x_node_x_widgets` for all nodegroups
The assumption is that this entry always exists. Widgets fetch settings based on this assumption. We should investigate, and potentially update and export//reimport the model to solve this issue. It may be a deeper issue with the non-localized-string datatype

While this issue persists, the `Concept Images` section of the Concept report shows infinite spinners for the labels of uploaded images.

johnatawnclementawn:
Resolves #293

Note: diff includes expanded serialization of i18n fields that we were previously missing for v8 graphs


aarongundel (approved):
works great, thanks!

Issue #353, opened 2025-06-22
form reset is too rigid


aarongundel:
After discussion with @chrabyrd there's a need for greater flexibility with the form reset functionality in components; some editor forms need custom reset functionality that the ComponentEditor doesn't provide.   We can break this out in a similar way to how we handle save.

chrabyrd:
ressolved in #354

PR #354, opened 2025-06-23
add injected reset to optionally reset forms (#353)

aarongundel:


chrabyrd:
after a chat w/ @aarongundel we decided this is unneccesary

Issue #341, opened 2025-06-11
Scheme buttons in list page should have status and functionality placeholders


chrabyrd:
<img width="527" alt="Image" src="https://github.com/user-attachments/assets/9cb199db-e66f-48a2-8368-86fb4337ef8c" />

PR #390, opened 2025-06-27
restyles scheme and adds menu placeholders (#341)

aarongundel:
scrolling is broken for the menubar here.  Otherwise this should be fairly close.


chrabyrd (changes requested):
I'm not sure if it's my local or not, but this is what I'm seeing. Could this match the mockup a bit closer?
Screenshot 2025-06-27 at 4 52 55 PM

chrabyrd:
closing as unplanned

PR #418, opened 2025-07-01
adds various QoL fixes

chrabyrd:


jacobtylerwalls (approved):
concept builder changes LGTM
***
Interesting question here about whether to allow these kinds of deletes. When I built the concept tree I feel like the philosophy was that the tree would never be patchy. If we need to support multiple incomplete trees ....?

PR #421, opened 2025-07-01
Loosen arches pin

jacobtylerwalls:
I reproduced the install failure from the deploy job locally, and this resolves it.

~= seems better than >=, we're far away from using arches 8.1.

The alternative would be to change component lab and controlled lists to point to dev/8 also, but component lab supports 7.6 that wouldn't have been friendly.


chrabyrd (approved):
huh, TIL. lgtm 🚀

PR #422, opened 2025-07-03
updates entrypoint of container to increase available workers

aarongundel:
adds workers equal to count of vcpu + 1


jacobtylerwalls (approved):
Nice. I saw advice to start with 2x num_cores, but maybe we can start slow and assess.

PR #434, opened 2025-08-15
lingo builds

chrabyrd:


arches_lingo/src/arches_lingo/utils.ts
```diff
@@ -53,6 +59,66 @@ export function navigateToSchemeOrConcept(
     }
 }
 
+/* Port of rank_label in arches.app.utils.i18n python module */
+export const rankLabel = (
```
jacobtylerwalls, line 63: What do you think about just importing these from controlled lists?

chrabyrd: yep! sg2m, I'll roll that into the next iteration of this PR

chrabyrd (commented):
~~This still needs to update to use the latest endpoints from querysets, so save is still borked until then. But you should be able to get Lingo up and running and the editor working for everything but save.~~

chrabyrd:
currently failing builds until https://github.com/archesproject/arches-controlled-lists/pull/156 lands

arches_lingo/src/arches_lingo/components/widgets/ConceptResourceSelectWidget/ConceptResourceSelectWidget.vue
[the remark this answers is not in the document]
chrabyrd: good Q!

_maybe_

This is based off an old version of the resource select widget, with some changes applied. Essentially the choice dropdown in this component shows the hierarchical path ( similar to the basic search function ) . 

It's broken because it's based on an old pattern. But I could see how knowing the path of what you're selecting could useful && necessary ( eg pillbox )

robgaston (approved):
lgtm! thanks!!

PR #453, opened 2025-09-17
Upgrades lingo to core 8.1

aarongundel:


arches_lingo/settings.py
```diff
@@ -278,6 +284,15 @@ def get_optional_env_variable(var_name, default=None) -> str:
...
 
+
+# def show_toolbar(request):
+#     return True
+
+
+# DEBUG_TOOLBAR_CONFIG = {
+#     "SHOW_TOOLBAR_CALLBACK": show_toolbar,
+# }
+# UPDATE_ON_FETCH = True
```
chrabyrd, line 295: cruft?

chrabyrd (changes requested):

arches_lingo/settings.py
chrabyrd: chatted w/Aaron and this is cruft. Removing because I'm already here

chrabyrd (approved):
LGTM

PR #455, opened 2025-09-22
Genericize lingo resource headers

### LINKED ISSUE #442: Enable Scheme & Concept export
The exporter logic in `SchemeHeader` and `ConceptHeader` are essentially the same - they should be broken out into a single component that will eventually be wired up when we have export logic on the backend. I think we open a follow-up for this for now to unblock Dennis @robgaston / @chrabyrd

_Originally posted by @johnatawnclementawn in https://github.com/archesproject/arches-lingo/pull/432#discussion_r2334323694_

johnatawnclementawn:
partially resolves #442


chrabyrd (changes requested):
There's def still a good bit more to polish here but I get && appreciate the effort. A few things before I deep-dive into the code:

- `header/` is confusing, as this isn't the application header, it's a subheader of two pages.
- Generic components shouldn't be hardcoding conditionals. eg `partOfScheme`. That and the `{SCHEME OR CONCEPT}` phrasing show that the abstraction isn't quite there yet.

There seems to be difference enough between the two components that maybe they _shouldn't_ be abstracted. Assuming a polished abstraction I could see a slot for buttons, links, etc etc, and at that point what _is_ the abstraction except a few CSS rules?

Not saying it shouldn't happen but serious though should be given to whether this is worth it or not. Happy to chat if you'd like 👍

PR #473, opened 2025-10-15
Import lingo resources directly from skos

### LINKED ISSUE #466: Enable import of Lingo Resources from SKOS RDF xml

### LINKED ISSUE #472: How should we handle languages not present in Languages controlled list
The current behavior in the RDM for concept import via skos file is as follows: if a language is encountered that does not exist in the Languages table, a record is inserted for that language in the Languages table and then the import moves forward.

Since we are now managing Languages as a controlled list, during an import of Lingo data (either from the RDM or directly from a skos file), if a language is not present in the Languages controlled list, should we:
1. Create a new list item (status quo)
2. Stop import process and prompt administrator to add new item to the list

johnatawnclementawn:
Resolves #466
~~Resolves #472~~ Will need to follow up once we create a 'Language' datatype (backed up by the Languages table

### DIFF
--- a/arches_lingo/etl_modules/migrate_to_lingo.py
+++ b/arches_lingo/etl_modules/migrate_to_lingo.py
@@ -223,14 +223,9 @@ class LingoResourceImporter(BaseImportModule):
             or value["valuetype_id"] == "historyNote"
             or value["valuetype_id"] == "scopeNote"
         ):
-            if isScheme:
-                mock_tile["statement_content_n1"] = value["value"]
-                mock_tile["statement_type_n1"] = value["valuetype_id"]
-                mock_tile["statement_language_n1"] = value["language"]
-            else:
-                mock_tile["statement_content"] = value["value"]
-                mock_tile["statement_type"] = value["valuetype_id"]
-                mock_tile["statement_language"] = value["language"]
+            mock_tile["statement_content"] = value["value"]
+            mock_tile["statement_type"] = value["valuetype_id"]
+            mock_tile["statement_language"] = value["language"]
             return {"statement": mock_tile}
         pass
 
--- a/arches_lingo/pkg/graphs/resource_models/Scheme.json
+++ b/arches_lingo/pkg/graphs/resource_models/Scheme.json
@@ -91,7 +91,7 @@
                     "config": null,
                     "constraints": [],
                     "cssclass": "big",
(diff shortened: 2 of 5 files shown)


johnatawnclementawn:
The current iteration does not support the use of celery. To illustrate the issue, let's look at the current order of operations:

1. Front-end sends POST request to `ThesaurusImportView`
2. skosfile and overwrite options are read from request object
3. Minimal file validation done
4. skosfile is parsed by `SKOSReader`
5. Lingo's `SKOSReader` subclass is used to extract schemes & concepts into lists of aliased data dictionaries (aka mock tiles in `LingoResourceImporter`
6. The extracted schemes, concepts, and relations are passed to the instantiated `LingoResourceImporter`'s call of `write()`
7. The `LingoResourceImporter`'s `write` method determines whether to process the task synchronously or async

Here is where the issue is encountered if we're using celery. Since the schemes, concepts, and relations have already been extracted from the skosfile, we need to pass them to the instance of the celery task, which then instantiates a new `LingoResourceImporter` instance. However, these lists of dicts would need to be serialized (along with all of the other `LingoResourceImporter` class attributes), but downstream we rely on the mock tiles to contain django model instances - serialization would mean losing this convenience (and also just feels like the wrong solution).

A potential solution is reordering the operations so that the skosfile parsing is done within the context of the celery task. This however requires passing the xml BLOB to the celery task and configuring celery to accept that binary obj (in form of pickle). 

We will also need to wire up the Lingo notifications system. When the use of celery is resolved, we will need to use the notification system to alert the user that the task has kicked over to celery.

johnatawnclementawn:
> @johnatawnclementawn this looks good, but wondering about the failing tests. could you take a look and let me know what you think?

I cherry-picked down a [commit](https://github.com/archesproject/arches-lingo/pull/473/commits/b30d6f9238a02f62f2b8c73f1ff62786dc3312b2) from the export branch that standardized a few node aliases in the `Scheme` model with those in the `Concept` model. The prior [commit](https://github.com/archesproject/arches-lingo/pull/473/commits/b269a88c043c2441960e8d09d23cc1ac4d09f07e)  allows for Schemes to have prefLabels that are skos types, not just dcterms:title & standardizes that predicate/object mapping into a single method used for both scheme & concept import, but that relied on the node alias updates, so tests were failing because the schemes mapping had failed.

All that to say, tests pass now - should be g2g

PR #484, opened 2025-11-13
Export Thesauri to SKOS

johnatawnclementawn:
Requires: https://github.com/archesproject/arches-querysets/pull/135

### DIFF
--- a/arches_lingo/src/arches_lingo/api.ts
+++ b/arches_lingo/src/arches_lingo/api.ts
@@ -376,7 +376,6 @@ export const exportThesaurus = async (
 ) => {
     const formData = new FormData();
     formData.append("action", "start");
-    // formData.append("mode", "ui");
     // moduleId normally fetched in context of BDM, but needed to get throough ETLManagerView POST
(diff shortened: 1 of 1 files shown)


arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/ExportThesauri.vue
```diff
@@ -0,0 +1,243 @@
...
+
+const props = defineProps<{
+    resourceId: string;
+    resourceName: string | undefined;
+}>();
+
+const loading = ref(false);
+const visible = ref(true);
+
+const exportDepth = ref("complete");
```
johnatawnclementawn: This probably doesn't need to exist as the depth is derived from the root resource you choose and it was expressed that we did not envision the use case for exporting a single concept outside of the context of a hierarchy. Can you confirm though?

johnatawnclementawn:
In this iteration, you can download the exported file from the Bulk Data Manager or access the file in the file system. #489 will take care of displaying download links in the Lingo UI for users

PR #485, opened 2025-11-27
Search Updates

chrabyrd:
- Fixes search fuzziness
- Adds ordering interactions to basic search
- Removes unneeded UI for model-based filtering

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ requires-python = ">=3.11"
 dependencies = [
     "arches @ git+https://github.com/archesproject/arches@dev/8.1.x",
     "arches_component_lab @ git+https://github.com/archesproject/arches-component-lab@main",
-    "arches_querysets @ git+https://github.com/archesproject/arches-querysets@dev/1.1.x",
+    "arches_querysets @ git+https://github.com/archesproject/arches-querysets@dev/1.0.x",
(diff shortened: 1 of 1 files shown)


arches_lingo/src/arches_lingo/components/basic-search/SortAndFilterControls.vue
```diff
@@ -1,27 +1,30 @@
...
-const queryFilters = ref([]);
+const ORDER_MODE_UNSORTED = "unsorted";
+const ORDER_MODE_ALPHABETICAL = "alphabetical";
+const ORDER_MODE_REVERSE_ALPHABETICAL = "reverse-alphabetical";
 
 const sortOptions = [
-    { value: "Unsorted", label: $gettext("Unsorted") },
-    { value: "A to Z", label: $gettext("A to Z") },
-    { value: "Z to A", label: $gettext("Z to A") },
+    { value: ORDER_MODE_UNSORTED, label: $gettext("Unsorted") },
```
robgaston: I think the label "Unsorted" here may be confusing, particularly given the improvements here, maybe we should use something like this?:
```suggestion
    { value: ORDER_MODE_UNSORTED, label: $gettext("Best match") },
```

robgaston (approved):
lgtm - thanks!

Issue #492, opened 2025-12-17
Only one image should be allowed to be uploaded to digital_object_rdm_system -> content node


robgaston:
Right now, the UX is confusing as you are allowed to add multiple images to a single digital object and it's not clear to the end user the distinction between that and creating multiple digital objects.  Forcing a one-to-one relationship between images and digital objects should allow us to improve the UX there and make it clear which notes/names apply to which images.

Issue #491, opened 2025-12-17
Matched concepts should use a non-localized string (URI) instead of a resource instance data type


robgaston:
It's not a good assumption that matched concepts will be in the local system (or even in a Lingo system at all) so we should instead use non-localized strings for this node where a user can paste in a URI.

We may need to modify our import/export logic accordingly.

(a follow on to this might be to make quickly copying local concept URIs in lingo, either from the hierarchy or the search, possible as a convenience)

johnatawnclementawn:
I suggest using the URL datatype instead of a non-localized string because we can presumably rely on the datatype component that's already built out in component lab (rather than needing to roll our own flavor of non-localized string viewer)

PR #495, opened 2025-12-19
fixes scheme publication id & standardizes publication notes

robgaston:
@johnatawnclementawn we should try to maintain publication ids and notes during development as we apply any model changes required for #491 & #492


johnatawnclementawn (approved):
lgtm 👍🏼

PR #522, opened 2026-01-27
cleans up loaded models and model names

robgaston:


johnatawnclementawn (changes requested):
ope! can you rm `rdm` from `person_rdm_system` and `place_rdm_system` file names

PR #544, opened 2026-02-02
Update Lingo models to use language datatype

### LINKED ISSUE #472: How should we handle languages not present in Languages controlled list

johnatawnclementawn:
Resolves #472


chrabyrd (approved):
LGTM 👍

PR #546, opened 2026-02-03
adds concept type widget to header to replace stub

robgaston:


arches_lingo/src/arches_lingo/components/concept/ConceptHeader.vue
```diff
@@ -230,13 +263,19 @@ function extractConceptHeaderData(concept: ResourceInstanceResult) {
...
                 </h2>
                 <div class="card flex justify-center">
-                    <Select
-                        v-model="conceptType"
-                        :options="ctype"
-                        option-label="name"
-                        placeholder="Concept"
-                        checkmark
-                        :highlight-on-select="false"
+                    <GenericWidget
```
robgaston: @chrabyrd I updated the model to assign a default value to this node, but my usage of `GenericWidget` here does not seem to pick it up.  Do you perhaps know why that would be the case? I wasn't able to see any issues while debugging, but the default value never shows up in the component.

A side note: we should assign a tile with the default values for the `type` tile on creation of a concept.  That's a follow-on I might ask for your help on as well unless it's easy to tack on here (if so, how should we do that?).

Thanks!

PR #549, opened 2026-02-09
Handle URI generation

chrabyrd:
Handles URI generation for schemes and concepts. Tightens logic around identifiers.

Ensure the models and test data updates have taken:

```
python manage.py packages -o load_package -s arches_lingo/pkg -db
```
```
python manage.py loaddata tests/fixtures/data/FISH_Example_Thesauri.json 
```

### DIFF
--- a/arches_lingo/pkg/graphs/resource_models/Concept.json
+++ b/arches_lingo/pkg/graphs/resource_models/Concept.json
@@ -5517,7 +5517,7 @@
                     "isrequired": false,
                     "issearchable": true,
                     "istopnode": false,
-                    "name": "uri",
+                    "name": "URI",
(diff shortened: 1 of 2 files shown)


robgaston (changes requested):
I've noticed a few functional issues:
- generated identifiers should use the type "identifier" (https://vocab.getty.edu/aat/300404012), currently they are being assigned null.
- concept URIs are generated, but they are not showing up in the ConceptHeader
- ldentifiers and URIs are now loaded for concepts and schemes (#505) but those loaded with test data are not shown nor used in URI generation
- test data do not load currently due to changed scheme publication id
- scheme URI includes concept route and identifier template.  it's okay for us to have the user simply input the scheme URI manually, but it would need to be required for publication

PR #550, opened 2026-02-09
[POC] shift focus to first form field when spawning

johnatawnclementawn:
Prove out watching loading state on GenericWidget to shift focus after form has rendered


johnatawnclementawn:
To simplify things, I'm going to merge this into my 538 base branch... fewer PR's && this is no longer a POC

PR #521, opened 2026-01-23
Cbyrd/handle scheme lifecycle states

chrabyrd:
Ensure the models and test data updates have taken:

```
python manage.py packages -o load_package -s arches_lingo/pkg -db

python manage.py loaddata tests/fixtures/data/FISH_Thesauri_example_data_resources.json tests/fixtures/data/FISH_Thesauri_example_data_tiles.json
```

PR #566, opened 2026-02-19
#505 follow up

johnatawnclementawn:
- Fixes issue with RouterLink in ConceptMatch Viewer
- Updates ConceptHeader to display URI and Identifier(s)

~~@robgaston - Schemes aren't modeled with a URI nodegroup, should that be stricken from the `SchemeHeader`?~~


robgaston:
@chrabyrd Should we rework this PR to back out the work on the ConceptHeader here as it will likely conflict with your work on #521 and #549?

PR #567, opened 2026-02-19
Ensure language widget mapping is created during Lingo project initialization

johnatawnclementawn:
Follow up to #544


robgaston (approved):
lgtm!

Issue #551, opened 2026-02-10
Implement localization in frontend


robgaston:
Currently the values for `selectedLanguageKey` and `systemLanguageKey` are being hardcoded.  We should implement the language switcher (#47) and system language (#311) we should also consider the localization of model information (#347) and provide for localizing card/widget labels coming from the models (even if we don't have the model labels themselves localized in the graph json ATM).

PR #569, opened 2026-02-20
implements language selector and gettext translation

robgaston:
re: #551 

Implements language selector and translates static strings and scheme/concept labels.

[#551](https://github.com/archesproject/arches-lingo/issues/551#issuecomment-3937839902) includes de translations of static strings for testing

### DIFF
--- a/arches_lingo/src/arches_lingo/components/basic-search/SearchResult.vue
+++ b/arches_lingo/src/arches_lingo/components/basic-search/SearchResult.vue
@@ -1,18 +1,13 @@
 <script setup lang="ts">
-import { inject, ref } from "vue";
+import { storeToRefs } from "pinia";
 import { getItemLabel } from "@/arches_controlled_lists/utils.ts";
 import { getParentLabels, getConceptIcon } from "@/arches_lingo/utils.ts";
-import {
-    FALLBACK_LANGUAGE,
-    selectedLanguageKey,
-    systemLanguageKey,
-} from "@/arches_lingo/constants.ts";
+import { useLanguageStore } from "@/arches_lingo/stores/useLanguageStore.ts";
 
 import type { PropType } from "vue";
 import type { SearchResultItem } from "@/arches_lingo/types.ts";
 
-const selectedLanguage = inject(selectedLanguageKey, ref(FALLBACK_LANGUAGE));
-const systemLanguage = inject(systemLanguageKey, FALLBACK_LANGUAGE);
+const { selectedLanguage, systemLanguage } = storeToRefs(useLanguageStore());
 
 defineProps({
     searchResult: {
--- a/arches_lingo/src/arches_lingo/components/header/PageHeader/components/LanguageSelector.vue
+++ b/arches_lingo/src/arches_lingo/components/header/PageHeader/components/LanguageSelector.vue
@@ -1,45 +1,40 @@
 <script setup lang="ts">
-import { computed, inject, useTemplateRef, watchEffect } from "vue";
+import { computed, useTemplateRef, watchEffect } from "vue";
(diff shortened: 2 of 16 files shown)


arches_lingo/src/arches_lingo/components/basic-search/SearchResult.vue
```diff
@@ -38,7 +45,11 @@ defineProps({
         <div class="search-result-hierarchy">
             [
             {{
-                getParentLabels(searchResult.option, ENGLISH.code, ENGLISH.code)
+                getParentLabels(
```
chrabyrd: same as above

arches_lingo/src/arches_lingo/api.ts
```diff
@@ -339,6 +339,37 @@ export const fetchConceptRelationships = async (
...
+export const fetchI18nData = async (languageCode?: string) => {
+    const url = arches.urls.api_get_frontend_i18n_data;
+    const headers: Record<string, string> = {};
+    if (languageCode) {
+        headers["Accept-Language"] = languageCode;
+    }
+    const response = await fetch(url, { headers });
+    const parsed = await response.json();
+    if (!response.ok) throw new Error(parsed.message || response.statusText);
+    return parsed as {
```
chrabyrd: less generic than the `Language` type, so maybe ok inline here. If you imagine it'll be used elsewhere I'd say move it to types.ts

chrabyrd (changes requested):
Overall looks pretty good, I think there's larger discussion to be had around where things live and how languages are switched. Happy to chat anytime of course 👍 

IMO and easy win that will really pare down complexity here is that `create-vue-application` fetches the correct i18n for the selected language on each mount. So as long as the Django source of truth is changed we could just refresh the page.

also consider the TODO in create-vue-app:

```
    /**
     * TODO: cbyrd #10501 - we should add an event listener that will re-fetch i18n data
     * and rebuild the app when a specific event is fired from the LanguageSwitcher component.
    **/
```

arches_lingo/urls.py
```diff
@@ -82,7 +81,4 @@
 # Only handle i18n routing in active project. This will still handle the routes provided by Arches core and Arches applications,
 # but handling i18n routes in multiple places causes application errors.
 if settings.ROOT_URLCONF == __name__:
-    if settings.SHOW_LANGUAGE_SWITCH is True:
```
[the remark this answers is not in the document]
robgaston: added a comment, but this is removed as introducing language codes into routes was breaking front-end interactions, so instead relied on cookies

arches_lingo/src/arches_lingo/utils.ts
```diff
@@ -274,6 +274,60 @@ export function extractDescriptors(
...
+export function getAutonym(code: string, fallback: string): string {
+    try {
+        const name = new Intl.DisplayNames([code], { type: "language" }).of(
+            code,
+        );
+        if (name) {
+            return name.charAt(0).toUpperCase() + name.slice(1);
+        }
+    } catch {
+        // Intl.DisplayNames may not support every code.
```
chrabyrd: empty catch?

arches_lingo/src/arches_lingo/components/header/PageHeader/components/LanguageSelector.vue
chrabyrd: This looks good for if we need to just get something out the door. I'd say Language handling would be a fantastic reason to dip our foot in to pinia if we can manage it. That abstracts most of this functionality to a simple store ( with side effects encapsulated there )

chrabyrd (changes requested):
Overall looking good, and there's some interesting larger pattern conversation to have if you'd like ( eg Pinia for this )

PR #572, opened 2026-02-24
improves tree ux when a new concept is saved

### LINKED ISSUE #539: When adding a new concept, Lingo appears to update all scheme hierarchies, instead of just the scheme being edited
If this is so, it would seem to an opportunity to improve performance.  If not, the UX makes it look like all schemes are being updated because the hierarchy panel shows only skeletons when a new concept is saved with its first label or note

robgaston:
fixes #539 

leaves tree visible during refresh, uses progress bar to indicate loading during refresh


chrabyrd:
closing this PR for now, until it looks like we'll need it or something similar again 👍

PR #579, opened 2026-02-25
adds basic dashboard for lingo homepage

### LINKED ISSUE #578: Add homepage with dashboard
We should show a dashboard on the homepage instead of the schemes list we show currently.

It should include:
- count of concepts
- recent edits
- missing translations

this should be filterable by scheme and language.

robgaston:
fixes #578 

note: I think that it is fine to consider this dashboard a temporary solution until we can fully adopt arches-search at which point it can be replaced


arches_lingo/views/api/dashboard.py
chrabyrd: usually try to guide people for fat models / skinny views. Mind abstracting a good bit of this to a util, service, or elsewhere?

Issue #526, opened 2026-01-30
Update labels and widget in forms


dwuthrich:
Data entry forms such as the Concept Notes forms and others need to use more intuitive labeling for form fields and widgets more appropriate for the information being collected.

Example: as shown in the attached image, we use the label "Statement" when we mean "Note", and the single text field is inappropriate; instead we should use either a text area widget or rich text widget.  

Similarly, the "Statement Type" label in the form should be replaced with "Note Type".

This pattern appears in the Image form; which should also be updated to use clearer, more appropriate labels/widgets.  For the image form, we should use "Image Title" and "Image Description" instead of the current labels

<img width="5050" height="2718" alt="Image" src="https://github.com/user-attachments/assets/53be2113-117d-4fe3-9faf-f22b0b07f815" />

PR #583, opened 2026-02-26
updates widget labels

### LINKED ISSUE #534: Change Labels in Associated Concepts form
Update "Associated Concept Relation Validity Period Begin" and "Associated Concept Relation Validity Period End" to "Relationship Begin Date" and "Relationship End Date"

robgaston:
fixes #534 
re: #526

### DIFF
--- a/arches_lingo/src/arches_lingo/components/scheme/SchemeLabel/components/SchemeLabelViewer.vue
+++ b/arches_lingo/src/arches_lingo/components/scheme/SchemeLabel/components/SchemeLabelViewer.vue
@@ -30,9 +30,9 @@ const openEditor = inject<(componentName: string) => void>("openEditor");
 
 const metaStringLabel: MetaStringText = {
     deleteConfirm: $gettext("Are you sure you want to delete this label?"),
-    language: $gettext("Label Language"),
-    name: $gettext("Label Name"),
-    type: $gettext("Label Type"),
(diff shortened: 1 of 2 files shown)


arches_lingo/pkg/graphs/resource_models/Concept.json
johnatawnclementawn:
```suggestion
                        "label": "Note",
```

arches_lingo/pkg/graphs/resource_models/Scheme.json
```diff
@@ -882,7 +882,7 @@
                         "i18n_properties": [
                             "placeholder"
                         ],
-                        "label": "Statement",
+                        "label": "Notes",
```
johnatawnclementawn: This should be singular
```suggestion
                        "label": "Note",
```

johnatawnclementawn (changes requested):
A few comments:
- #526 talks about using a rich text widget - that probably requires some component lab work to enable rich text editing for non-localized strings, I just want to make sure we don't lose sight of that request if we close out that ticket
- Is `Statement` still okay for `Scheme Rights` or should that also be updated?

robgaston:
great points.  I updated the labels and moved #526 so it should be retained after merge.  We should do the rich text widget work in a follow up as you are correct that it will need some archesproject/arches-component-lab work

PR #589, opened 2026-02-27
adds user profile page

### LINKED ISSUE #588: User profile page should be inside Lingo
user profile page calls out to core arches, but should be inside of lingo ui

robgaston:
fixes #588

### DIFF
--- a/arches_lingo/src/arches_lingo/api.ts
+++ b/arches_lingo/src/arches_lingo/api.ts
@@ -47,7 +47,9 @@ export const fetchUser = async () => {
 };
 
 export const fetchUserProfile = async (): Promise<User> => {
-    const response = await fetch(generateArchesURL("arches_lingo:api_lingo_user_profile"));
+    const response = await fetch(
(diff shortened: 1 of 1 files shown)


arches_lingo/views/api/user_profile.py
```diff
@@ -0,0 +1,161 @@
...
+                title=_("Invalid request"),
+                message=_("Request body must be valid JSON."),
+                status=400,
+            )
+
+        old_password = data.get("old_password", "")
+        new_password = data.get("new_password", "")
+        new_password2 = data.get("new_password2", "")
+
+        if not user.check_password(old_password):
```
chrabyrd: question: if the user is already authenticated, do we _want_ to force them to re-enter their old password to change it? Not a super strong opinion here either way

robgaston: I think so, but I'm open.  seems to me like the safe thing to do is require entry of the old password to change in case, for example, a user leaves a session open on a browser that can be accessed by others

PR #600, opened 2026-03-03
shows languages in scheme header

### LINKED ISSUE #599: Show languages user (with label counts) in scheme header
Currently the scheme header has hardcoded languages. These should be replaced with a calculated set with label counts.

robgaston:
fixes #599


johnatawnclementawn (approved):
This looks good! I was going to suggest just using the TileTrees that had already been fetched as part of `Concept = ResourceTileTree.get_tiles("concept")`, but hacking that out quickly resulted in 2x wait time, so I think what you have right now is good.

PR #601, opened 2026-03-03
Assigns concept type on creation

johnatawnclementawn:
Supersedes #568


robgaston (approved):
lgtm! this is ready for merge in my mind, but can we also update the fixtures to assign concept as the type to all test data please?

PR #617, opened 2026-03-07
allows anonymous/readonly access to lingo

robgaston:
also, replaces usage of "RDM Administrator" group w/ new "Lingo Editor" group

### DIFF
--- a/arches_lingo/src/arches_lingo/ArchesLingo.vue
+++ b/arches_lingo/src/arches_lingo/ArchesLingo.vue
@@ -50,6 +50,7 @@ onMounted(function () {
 watchEffect(() => {
     router.beforeEach(async (to, _from, next) => {
         if (to.name === routeNames.login) {
+            shouldShowHierarchy.value = false;
             next();
             return;
         }


arches_lingo/views/api/user_profile.py
```diff
@@ -14,6 +14,31 @@
...
+    def get(self, request):
+        user = request.user
+        is_anonymous = not user.is_authenticated or user.username == "anonymous"
+
+        return JSONResponse(
+            {
+                "username": user.username,
+                "first_name": getattr(user, "first_name", ""),
+                "last_name": getattr(user, "last_name", ""),
+                "is_editor": is_lingo_editor(user),
```
chrabyrd: nit to prevent headache with potential future complexity: `is_lingo_editor`

arches_lingo/src/arches_lingo/components/generic/EditLog/EditLog.vue
```diff
@@ -282,7 +284,7 @@ async function acceptRevert() {
                             {{ formatEditTypeLabel(edit) }}
                         </span>
                         <Button
-                            v-if="canRevert(edit)"
+                            v-if="isEditor && canRevert(edit)"
```
chrabyrd: nit: would instead just update `canRevert` to keep things simple

robgaston:
i got that UI issue cleared up in the latest.  I think that the dashboard is okay for  anon access, but will run that by the stakeholders

PR #620, opened 2026-03-09
Adds sources and contributors pages

robgaston:
ResourceListEditor is likely useful beyond Lingo, but thinking that can be a future change when the need emerges.

### DIFF
--- a/arches_lingo/pkg/graphs/resource_models/group_system.json
+++ b/arches_lingo/pkg/graphs/resource_models/group_system.json
@@ -673,7 +673,7 @@
                                 "string_template": ""
                             },
                             "name": {
-                                "nodegroup_id": "5bc65fd2-bb18-11ea-85a6-3af9d3b32b71",
+                                "nodegroup_id": "de76dbdc-11e5-11ef-9493-0a58a9feac02",
                                 "string_template": "<name_content>"
                             }
                         },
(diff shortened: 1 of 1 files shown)


arches_lingo/src/arches_lingo/components/generic/ResourceListEditor/ResourceListEditor.vue
```diff
@@ -0,0 +1,421 @@
...
+.list-header {
+    display: flex;
+    flex-direction: column;
+}
+
+.list-header-toolbar {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    min-height: 3rem;
```
chrabyrd: same with min-anything. Sometimes it's necessary but often it just creates headache

arches_lingo/src/arches_lingo/components/generic/ResourceListEditor/ResourceListEditor.vue
```diff
@@ -0,0 +1,427 @@
...
+                            </Button>
+                        </div>
+                    </div>
+                    <div class="editor-body">
+                        <slot
+                            name="editor"
+                            :resource="selectedResource"
+                            :is-creating-new="isCreatingNew"
+                        />
+                    </div>
```
johnatawnclementawn: I know the GenericCard has `Save` and `Cancel` buttons built in, but I think it may be easier to have consistent styling with the `ComponentEditor` if we borrow it's header & footer.

I'm also seeing a duplication of the word `Name` - likely the GenericCard is rendering the nodegroup's label + the GenericCardEditor is rendering the node's label.

chrabyrd (approved):
lgtm 👍

Issue #476, opened 2025-10-22
Hierarchical Position: When adding new Parent limit selection to concepts in the same scheme


Philbox:
like issue #474 the resource selection needs to be limited to the concepts in the current scheme being edited when adding a new parent as only concepts in the same scheme can be parents

<img width="1784" height="1147" alt="Image" src="https://github.com/user-attachments/assets/f454b2cb-b1d7-4e72-9a0d-8f29c7ffcd5f" />

PR #627, opened 2026-03-11
Get scheme from resource store, #476

njkim:
Limit the concept selections within the current scheme by getting scheme id.


robgaston (approved):
lgtm!

PR #628, opened 2026-03-11
adds lifecycle filter to hierarchy view

robgaston:


chrabyrd:
this is great! Mind if it gets put on ice until the delete concept stuff lands? As part of it I moved concept fetching to a pinia store and it makes the app more performant

robgaston:
yup, sounds good!

PR #630, opened 2026-03-11
adds sorting by label in hierarchy and top concepts

### LINKED ISSUE #373: Sort concepts alphabetically
Currently the hierarchy displays the concepts in an apparently random order (assume it's ordered by UUID or URI). See screenshot

It would be helpful if concepts could be displayed alphabetically by preflabel

Image

robgaston:
fixes #373


robgaston:
I centralized the sort logic and added sorting for schemes.  I'd argue that the other stuff should be done in follow up but let me know what you think

PR #633, opened 2026-03-12
Remove current concept from associated concept selection, #558

### LINKED ISSUE #558: Associated Concepts: Associating a concept to itself
When adding an associated concept it is possible to add a concept to itself. This should not be possible

Image

njkim:


robgaston (approved):
lgtm

PR #641, opened 2026-03-12
only requests scheme data for existing resources

robgaston:
follow up to #627


njkim (approved):
Good catch. Looks good.

PR #642, opened 2026-03-13
removes duplicate model json

robgaston:
#544 introduced duplicate model json for Concept model


johnatawnclementawn (approved):
whoops! lgtm

PR #646, opened 2026-03-13
adds top concept of section for top concepts

robgaston:
also, removes schemes from hierarchy forms.

top concepts can now no longer add hierarchical parents, and schemes can no longer be assigned as hierarchical parents (top_concept_of) to concepts. this effectively makes top concept status immutable, which matches requirements described today by @Philbox


njkim (approved):
lgtm

PR #649, opened 2026-03-13
uses label type uris for comparison over labels

### LINKED ISSUE #289: Use List Item URI for concept valuetype and language lookups
Not blocking getting this fix in to resolve the regression, but I think it's worth thinking about avoiding hard-coded values by comparing URI's (or some other known constant) - I don't think there's anything stopping a Lingo admin from editing the prefLabel for `preferred label` in the Controlled List Manager - which could mean that tiles are saved with value that is not going to pass this lookup

_Originally posted by @johnatawnclementawn in https://github.com/archesproject/arches-lingo/pull/288#discussion_r2049116639_

robgaston:
fixes #289


johnatawnclementawn (approved):
looks good 👍🏼

PR #657, opened 2026-03-17
Assign correct lifecycle state of schemes & concepts during import

### LINKED ISSUE #653: Imported schemes & concepts should have `Draft` lifecycle state
Initially the requirement was that if all of the concepts & schemes in an incoming thesaurus had URI's and Identifiers, the scheme could be considered `Published`. If the thesaurus had some concepts with URIs and Identifiers, the thesaurus should be in `Edit` mode, with the concepts not having URIs and identifiers being in `Draft`.

After talking with @robgaston, it seems more likely that folks will be using the import tooling to achieve two things: 1. migrate from the RDM manage their hierarchies wholesale 2. Create composite vocabularies. In both scenarios, the identifiers/URIs that are coming over are really externally matched concepts (or in the first scenario just stand-ins that were created by Arches in the first place and therefore do not need to be imported). Therefore, during an import, we will save any identifiers/URIs that do not match the conceptid (to avoid persisting the RDM generated URI) into a matched concept tile. Then, if necessary, as a precursor to publishing their thesaurus, users can programmatically or manually ETL those identifiers from matched_concepts to the URI and Identifier records.

johnatawnclementawn:
Resolves #653

### DIFF
--- a/arches_lingo/utils/skos.py
+++ b/arches_lingo/utils/skos.py
@@ -147,15 +147,15 @@ class SKOSReader(SKOSReader):
                         new_concept["tile_data"].append(mock_tile)
 
                     for predicate, object in graph.predicate_objects(concept):
-                        # Cast dcterms:description to scopeNote (same behavior in RDM)
                         if predicate == DCTERMS.description:
+                            # Cast dcterms:description to scopeNote (same behavior in RDM)
                             predicate_str = "scopeNote"
-
-                        predicate_str = (
-                            predicate.replace(ARCHES, "")
-                            .replace(SKOS, "")
-                            .replace(DCTERMS, "")
-                        )
+                        else:
+                            predicate_str = (
+                                predicate.replace(ARCHES, "")
+                                .replace(SKOS, "")
+                                .replace(DCTERMS, "")
+                            )
 
                         if predicate in [
                             SKOS.broader,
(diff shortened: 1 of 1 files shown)


arches_lingo/const.py
```diff
@@ -129,6 +129,11 @@
 # status (lifecycle/status)
 STATUS_NODEGROUP = "152321b6-0f5e-11ef-9493-0a58a9feac02"
 
+# Lifecycle states
```
robgaston: these appear to be unused, maybe an artifact of work here

robgaston (changes requested):
This mostly looks good (small nit), but I tested import and it doesn't seem to be working, at least with this file:
[AAT_scientific_concepts.xml](https://github.com/user-attachments/files/26129908/AAT_scientific_concepts.xml)
I've successfully imported this in the past.  it appears to be creating the concepts but not the scheme, but i didn't look deeply into it

johnatawnclementawn:
> most other imports I tried were successful, except for when importing this file: [Historic_Landscape_Characterization_20201102.xml](https://github.com/user-attachments/files/26130081/Historic_Landscape_Characterization_20201102.xml) i'm observing this error in celery:

Unfortunately, I am unable to repro this. Could you try again with the latest and lmk?

PR #675, opened 2026-03-19
catches missing gettext strings

robgaston:
found some edge cases leading to strings not being included in .po files

### DIFF
--- a/arches_lingo/src/arches_lingo/components/advanced-search/FacetRow.vue
+++ b/arches_lingo/src/arches_lingo/components/advanced-search/FacetRow.vue
@@ -74,6 +74,20 @@ const hierarchyDirections = [
     { label: $gettext("Narrower (child of)"), value: "narrower" },
 ];
 
+const searchTextPlaceholder = $gettext("Search text...");
+const noConceptsFoundMessage = $gettext("No concepts found");
+const searchConceptsFilterPlaceholder = $gettext("Search concepts...");
+const selectConceptPlaceholder = $gettext("Select concept...");
+const anyTypePlaceholder = $gettext("Any type");
+const anyLanguagePlaceholder = $gettext("Any language");
+const selectLanguagePlaceholder = $gettext("Select language");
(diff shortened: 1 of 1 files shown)


arches_lingo/src/arches_lingo/components/dashboard/RecentActivityTable.vue
```diff
@@ -127,95 +127,98 @@ function getDisplayName(item: DashboardActivityItem): string {
...
+                >
+                    <template #body="slotProps">
+                        <Tag
+                            :value="
+                                slotProps.data.resource_type === 'scheme'
+                                    ? $gettext('Scheme')
+                                    : $gettext('Concept')
+                            "
+                            :severity="
+                                slotProps.data.resource_type === 'scheme'
```
chrabyrd: same, also SCHEME, SUCCESS, and INFO are already declared in constants.ts I think

arches_lingo/src/arches_lingo/components/advanced-search/FacetRow.vue
```diff
@@ -245,192 +245,212 @@ function toggleNegated() {
...
+                "
+            />
+        </template>
+
+        <template>
+            <Button
+                icon="pi pi-times"
+                severity="danger"
+                text
+                rounded
```
chrabyrd: nit: explicit props

chrabyrd (commented):
few minor things, also why move the set of v-if's to `<template>`s? Seems unnecessary?

robgaston:
@chrabyrd oh, the template tags around are what allow gettext to properly retrieve the strings. the parser used by gettext handles tags that conflict with HTML5 tag names (aside from case) and this causes issues with Select/select tags. the parser converts the first Select to a select and enters a mode that ignore subsequent unknown elements.  thus the need to wrap Selects  in individual template tags as there is an exception carved out for handling template tags. I moved the conditionals off of these template tags as they are just there as a workaround for this behavior and that seemed like a better approach

robgaston:
oh good call! will rework

PR #683, opened 2026-03-24
Hide lifecycle buttons from non-editor

johnatawnclementawn:


robgaston (approved):
lgtm! thanks for catching this

PR #703, opened 2026-04-06
Import fixes for AAT data load

robgaston:
SKOS topConceptOf/language handling, ETL batched staging inserts

### DIFF
--- a/arches_lingo/etl_modules/migrate_to_lingo.py
+++ b/arches_lingo/etl_modules/migrate_to_lingo.py
@@ -354,10 +354,6 @@ class LingoResourceImporter(BaseImportModule):
             if tiles_to_load:
                 LoadStaging.objects.bulk_create(tiles_to_load)
 
-        # cursor.execute(
-        #     """CALL __arches_check_tile_cardinality_violation_for_load(%s)""",
-        #     [self.loadid],
-        # )
         cursor.execute(
             """
                 INSERT INTO load_errors (type, source, error, loadid, nodegroupid)
--- a/arches_lingo/etl_modules/migrate_to_lingo.py
+++ b/arches_lingo/etl_modules/migrate_to_lingo.py
@@ -47,6 +47,12 @@ details = {
     "helptemplate": "migrate-to-lingo-help",
 }
 
+# Number of concepts to process per bulk_create batch in populate_staging_table.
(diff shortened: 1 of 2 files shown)


arches_lingo/utils/skos.py
```diff
@@ -157,7 +155,23 @@ def extract_concepts_from_skos_for_lingo_import(
...
+                            )
+                            top_concept_mock_tile = LingoResourceImporter.create_mock_tile_from_relationship(
+                                {
+                                    "resourceId": scheme_of_top_concept_pk,
+                                    "node_alias": "top_concept_of",
+                                    "nodegroup_alias": "top_concept_of",
+                                    "resourceinstanceid": concept_pk,
+                                }
+                            )
+                            new_concept["tile_data"].append(top_concept_mock_tile)
```
johnatawnclementawn, line 179: Good catch re. valid skos using `hasTopConcept` and`topConceptOf` - I think we should put a guard here in case a thesaurus defines the relationship both ways (I can't remember exactly where, but I recall seeing this somewhere), since that would cause cardinality issues. Maybe it's enough to check if `new_concept` already has a `top_concept_of` mock_tile in `aliased_data`, or mark the `new_concept` as a top_concept during the `if top_concept_mock_tiles.get(concept_pk):` block and skip this if true (my preference for latter to avoid unnecessary iteration)

arches_lingo/etl_modules/migrate_to_lingo.py
```diff
@@ -298,45 +304,60 @@ def create_mock_tile_from_relationship(relationship):
...
+                        )
+                    )
+            if tiles_to_load:
+                LoadStaging.objects.bulk_create(tiles_to_load)
 
-        cursor.execute(
-            """CALL __arches_check_tile_cardinality_violation_for_load(%s)""",
-            [self.loadid],
-        )
+        # cursor.execute(
```
johnatawnclementawn: You're good to just remove this - base import module will take care of it on Cyrus's branch

johnatawnclementawn (commented):
This looks good overall - really just one thing for consideration

arches_lingo/utils/skos.py
robgaston: good catch! i've added this guard

johnatawnclementawn (approved):
LGTM 👍

PR #704, opened 2026-04-06
Search performance optimization for large scale datasets

robgaston:
raw SQL SearchResultSet with GIN trigram index, advanced search QuerySet composition, batch search result enrichment


chrabyrd (commented):
code looks fine overall -- how big is the win here? inlining SQL is harder to maintain but I get it if there's a big necessary win using it

robgaston:
totally agree. with AAT loaded, simple term searches were really slow (~20s) but these changes got them to ~1s, so the win was quite dramatic

chrabyrd (approved):
LGTM!

chrabyrd:
> totally agree. with AAT loaded, simple term searches were really slow (~20s) but these changes got them to ~1s, so the win was quite dramatic

woof yeah that'll do it 🚀

PR #715, opened 2026-04-07
adds all results to a set in advanced search

### LINKED ISSUE #710: Advanced Search: add ability to add entire result set to saved set
Select All -> Add to set should result in all results being added to the set unless user changes individual selections.  currently it only adds the current page.

robgaston:
fixes #710

### DIFF
--- a/arches_lingo/src/arches_lingo/components/advanced-search/ConceptSets.vue
+++ b/arches_lingo/src/arches_lingo/components/advanced-search/ConceptSets.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, watch, withDefaults } from "vue";
+import { ref, watch } from "vue";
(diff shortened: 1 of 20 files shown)


arches_lingo/src/arches_lingo/components/advanced-search/ConceptSets.vue
```diff
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, watch } from "vue";
+import { ref, watch, withDefaults } from "vue";
```
chrabyrd: `withDefaults` is a compiler macro now and we get a warning when building the bundle and it's imported like this

chrabyrd (changes requested):
looks good, just the one nit to stop the compiler warnings

chrabyrd (approved):
LGTM!

PR #717, opened 2026-04-09
improves scheme header display for schemes with many languages

robgaston:


arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -664,6 +706,50 @@ h2 > span {
...
+}
+
+.language-section-header {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+}
+
+.language-chip-container--expanded {
+    max-height: 10rem;
```
chrabyrd: this causes some minor UI issues at higher resolutions, works fine if it's removed:

Image

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -507,21 +518,51 @@ function onLifecycleStateChange(
...
                         >
-                            {{ entry.language }} ({{ entry.code }}):
-                            {{ entry.count }}
-                        </span>
+                            <Skeleton
+                                v-if="labelCountsLoading"
+                                width="12rem"
+                                height="2rem"
+                            />
+                            <span
```
chrabyrd, line 576: I know this is just a move but I think the i18n is better served like this:

```
const visibleLabelChips = computed(() =>
    visibleLabelCounts.value.map((entry) => ({
        key: entry.code,
        label: interpolate(
            $gettext("%{language} (%{code}): %{count}"),
            { language: entry.language, code: entry.code, count: entry.count },
            true,
        ),
    })),
);
```

So that translators will have a little context if anything needs to shift

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -507,21 +518,51 @@ function onLifecycleStateChange(
...
+                                @click="showAllLanguages = !showAllLanguages"
+                            >
+                                <template v-if="showAllLanguages">
+                                    <i class="pi pi-chevron-up"></i>
+                                    {{ $gettext("Show less") }}
+                                </template>
+                                <template v-else>
+                                    <i class="pi pi-chevron-down"></i>
+                                    +{{ hiddenLanguageCount }}
+                                    {{ $gettext("more") }}
```
chrabyrd: same here: 

```
const showMoreLabel = computed(() =>
    interpolate(
        $gettext("+%{count} more"),
        { count: hiddenLanguageCount.value },
    ),
);
```

also don't forget to wrap in a span or something 😄

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -507,21 +518,51 @@ function onLifecycleStateChange(
...
+                            <button
+                                v-if="hiddenLanguageCount > 0"
+                                type="button"
+                                class="language-expand-toggle"
+                                :aria-expanded="showAllLanguages"
+                                @click="showAllLanguages = !showAllLanguages"
+                            >
+                                <template v-if="showAllLanguages">
+                                    <i class="pi pi-chevron-up"></i>
+                                    {{ $gettext("Show less") }}
```
chrabyrd: same, wrap in an html element

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -507,21 +518,51 @@ function onLifecycleStateChange(
...
-                            v-for="entry in labelCounts"
-                            v-else
-                            :key="entry.code"
-                            class="scheme-language"
+                    <div class="language-chip-wrapper">
+                        <div class="language-section-header">
+                            <span class="header-item-label">{{
+                                $gettext("Languages:")
+                            }}</span>
+                            <button
```
chrabyrd: primevue button?

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -93,6 +93,20 @@ const label = ref<Label>();
...
+const LANGUAGE_CHIPS_TRUNCATE_COUNT = 8;
+
+const visibleLabelCounts = computed(() =>
+    showAllLanguages.value
+        ? labelCounts.value
+        : labelCounts.value.slice(0, LANGUAGE_CHIPS_TRUNCATE_COUNT),
+);
+
+const hiddenLanguageCount = computed(() =>
+    Math.max(0, labelCounts.value.length - LANGUAGE_CHIPS_TRUNCATE_COUNT),
```
chrabyrd: guarding against negative numbers here? I don't think it'd be possible to hit this

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
```diff
@@ -93,6 +93,20 @@ const label = ref<Label>();
 const data = ref<SchemeHeader>();
 const labelCounts = ref<LanguageLabelCount[]>([]);
 const labelCountsLoading = ref(false);
+const showAllLanguages = ref(false);
+
+const LANGUAGE_CHIPS_TRUNCATE_COUNT = 8;
+
+const visibleLabelCounts = computed(() =>
+    showAllLanguages.value
```
chrabyrd: nit: no multi-line ternaries

chrabyrd (changes requested):
looks good overall! just some pretty minor stuff with the only important part being the i18n changes

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/SchemeHeader.vue
robgaston: this is awkward at maybe a dozen languages, but AAT (as an example) has 168 languages making the header extremely tall without the max height in place.  maybe this is okay, given you can collapse the language list, but it renders the page content inaccessible even at normal window sizes as the header quickly scrolls and there is no way to scroll past it. i'll consider alternatives...
Screenshot 2026-04-14 at 2 37 27 PM

robgaston: seems okay to just rely on expand/collapse here, i suppose

chrabyrd: oooo nice! Good call out

chrabyrd (approved):
LGTM!

PR #725, opened 2026-04-14
adds script to update concept types and updates list

robgaston:


arches_lingo/management/commands/update_aat_concept_types.py
```diff
@@ -0,0 +1,283 @@
...
+  - **guide term** — the ~1 800 non-indexable organisational terms displayed
+    with angle brackets (e.g. ``<complexes by function>``).  Detected by the
+    English label starting with ``<`` and ending with ``>``.
+
+All other concepts retain the default "concept" type and are not touched.
+
+Prerequisites:
+  1. The AAT SKOS data has been imported via the normal lingo import:
+     ``python manage.py packages -o import_lingo_resources -s getty_aat_skos.xml``
+  2. The migration ``0013_add_aat_concept_types`` has been applied so that
```
johnatawnclementawn, line 30: I'm not seeing this migration, is that supposed to be included in this PR?

robgaston: sorry, i didn't commit the migration because it was just inserting the records from the list that I included. I didn't think it really made sense to introduce a migration here, though i did use it for early local testing.  I think you can just update the list from the XML and won't need the migration, but let me know if you want me to add it here for your testing

PR #729, opened 2026-04-16
enhancements and fixes to advanced search

robgaston:
allows for cascading hierarchical searches, enhancements to concept dropdowns, fixes for loading saved searches


chrabyrd (commented):
code looks good, 2 functional nits while we're here:

`view ->` should probably open in a new tab, since navigating back to the advanced search clears the query

after typing in the search bar, I'd expect hitting the enter key would execute the search

robgaston:
@chrabyrd good call. added those in the latest

chrabyrd (approved):
LGTM!

Issue #695, opened 2026-03-27
Retiring a vocabulary: Needs a warning and to be reversible


Philbox:
When retiring a vocabulary a pop up box should appear asking if the user is sure they want to retire a vocabulary

Retirement should be reversible

robgaston:
fixed in #731

PR #731, opened 2026-04-17
harden lifecycle retirement #695

chrabyrd:


robgaston (approved):
lgtm! perhaps we should test(s) for unretire, but not going to block on this

chrabyrd:
> lgtm! perhaps we should test(s) for unretire, but not going to block on this

added those tests 😄

PR #732, opened 2026-04-17
Improves report ux large datasets and adds narrower concepts section

robgaston:
@chrabyrd heads up - I know we discussed the narrower concepts section as something you would work on, but when working on handling large datasets better in reports that this section would need the same treatment here as there may be hundreds of children, so I threw it in here as it was pretty trivial


chrabyrd (approved):
LGTM!

PR #734, opened 2026-04-17
Cbyrd/update scheme and concept buttons 2

chrabyrd:


robgaston:
This looks good to me, but I'll run it past @dwuthrich for a gut check on the new styles.

robgaston (changes requested):
I notice that the edit/delete buttons for the Hierarchical Position section are still using the old style. could you please update them to be consistent with the new style?

PR #735, opened 2026-04-20
Cbyrd/add scheme and concept counter

chrabyrd:


robgaston (changes requested):
this looks good to me.  a minor style fix: the information icon next to "URI Template" is getting cut off on the sides.  can you fix this style? also, I sort of think it would be better to show the icon to the right of the label

PR #744, opened 2026-05-06
Improves image upload and display

robgaston:


arches_lingo/src/arches_lingo/components/concept/ConceptImages/components/editorState.ts
johnatawnclementawn, line 1: Do you think using a composable for this is a little over-powered? For state like this that's only going to be shared across 2 sibling components, I think it would be cleaner to define  `targetDigitalObjectResourceInstanceId` in `ConceptImages` and share it with the children as a prop && use emits to update it when needed.

johnatawnclementawn (commented):
One comment for your consideration

arches_lingo/src/arches_lingo/components/concept/ConceptImages/components/ConceptImagesViewer.vue
```diff
@@ -84,6 +83,35 @@ const resources = ref<DigitalObjectInstance[]>();
 const { $gettext } = useGettext();
 const confirm = useConfirm();
 
+function getFileUrl(originalUrl: string): string {
```
chrabyrd: Pretty generic, think it should go in utils.ts?

arches_lingo/src/arches_lingo/components/concept/ConceptImages/components/ConceptImagesViewer.vue
```diff
@@ -84,6 +83,35 @@ const resources = ref<DigitalObjectInstance[]>();
...
+    const contentData = resource.aliased_data.content?.aliased_data
+        .content as unknown as FileListValue | undefined;
+    const fileReference = contentData?.node_value?.[0];
+    if (fileReference?.url) {
+        return getFileUrl(fileReference.url);
+    }
+    return undefined;
+}
+
+function getImageAlt(resource: DigitalObjectInstance): string {
```
chrabyrd: these I'm less about where these should live, but if they live here mind moving them below `onMounted`?

chrabyrd (changes requested):
overall looks good! And I echo @johnatawnclementawn 's comment with the caveat that if we _need_ module-level state management for this for some reason, just use `pinia` since that pattern is being used already 👍

arches_lingo/src/arches_lingo/components/concept/ConceptImages/components/editorState.ts
robgaston: moved to a store as per @chrabyrd

arches_lingo/src/arches_lingo/components/concept/ConceptImages/components/ConceptImagesViewer.vue
robgaston: good idea, done

chrabyrd (approved):
LGTM!

PR #745, opened 2026-05-06
adds multiline config for notes and image descriptions

robgaston:
depends on archesproject/arches-component-lab#283


johnatawnclementawn (commented):
Should this be something that is memorialized in the graph JSON rather than being hard-coded on the front-end?

chrabyrd (approved):
I updated this to match the `renderContext` pattern in component lab. LGTM!

Also if I've jumped the gun on this, sorry and feel free to roll back my commit 😅

PR #751, opened 2026-06-03
Updates usage of rdflib && supports python 3.14

johnatawnclementawn:
Requires https://github.com/archesproject/arches/pull/12790


chrabyrd (changes requested):
did some frontend / webpack stuff leak into this, or is that part of the rdflib update?

johnatawnclementawn:
> did some frontend / webpack stuff leak into this, or is that part of the rdflib update?

I ran `updateproject` as part of ensuring that the `Arches` dep was up-to-date with the latest from 8.2.

chrabyrd (approved):
makes sense -- LGTM!

PR #752, opened 2026-06-05
adds scheme attribution

chrabyrd:


arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/SchemeAttributionField.vue
```diff
@@ -0,0 +1,140 @@
...
+
+<template>
+    <div class="header-item">
+        <span class="header-item-label">
+            {{ $gettext("Attribution:") }}
+        </span>
+
+        <template v-if="isEditingAttribution">
+            <div class="attribution-input-wrapper">
+                <InputText
```
johnatawnclementawn: Maybe we use `TextArea` or a rich text editor? I believe these attributions could be 1+ paragraph(s)

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/SchemeAttributionField.vue
```diff
@@ -0,0 +1,140 @@
...
+                size="small"
+                :rounded="true"
+                :aria-label="$gettext('Cancel')"
+                :disabled="isSavingAttribution"
+                @click="cancelEditingAttribution"
+            />
+        </template>
+
+        <template v-else>
+            <span class="header-item-value">
```
johnatawnclementawn: I think the attribution should wrap instead of `overflow: hidden` or at least expandable so we can see the whole attribution. Right now if there is overflow and your usertype is not `lingo editor`, you can't read the whole attribution.

arches_lingo/models.py
```diff
@@ -34,6 +34,20 @@ class Meta:
...
+        db_column="scheme_resource_instance_id",
+        on_delete=models.CASCADE,
+        related_name="scheme_attribution",
+    )
+    attribution = models.TextField(blank=True, default="")
+
+    class Meta:
+        db_table = "scheme_attributions"
+
+
```
johnatawnclementawn, line 86: For this model (and the other `Concept` or `Scheme` specific addendum models), could we implement `clean` & `save` methods that enforce that the related resource is an instance of a specific Arches graph:


```suggestion
from django.core.exceptions import ValidationError
from arches_lingo.const import CONCEPTS_GRAPH_ID, SCHEMES_GRAPH_ID

...

def clean(self):
      super().clean()
      if self.scheme and str(self.scheme.graphid) != str(SCHEMES_GRAPH_ID):
          raise ValidationError(
              {
                  "scheme": _(
                      "The related resource must be an instance of a scheme."
                  )
              }
          )

  def save(self, *args, **kwargs):
      self.full_clean()
      return super().save(*args, **kwargs)
```

johnatawnclementawn (commented):
Looking good - A few comments for your consideration

arches_lingo/models.py
chrabyrd: Good idea, I added that to all related models 👍

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/SchemeAttributionField.vue
```diff
@@ -0,0 +1,171 @@
...
+            />
+            <div class="attribution-input-wrapper">
+                <Textarea
+                    v-model="attributionDraft"
+                    :rows="3"
+                    :placeholder="
+                        $gettext('Enter attribution information here...')
+                    "
+                />
+            </div>
```
johnatawnclementawn:
```suggestion
             <div class="attribution-input-wrapper">
                <Textarea
                    v-model="attributionDraft"
                    :rows="3"
                    :placeholder="
                        $gettext('Enter attribution information here...')
                    "
                />
                <div class="attribution-edit-controls">
                    <Button
                        icon="pi pi-check"
                        variant="text"
                        severity="success"
                        size="small"
                        :rounded="true"
                        :aria-label="$gettext('Save attribution')"
                        :loading="isSavingAttribution"
                        @click="saveAttribution"
                    />
                    <Button
                        icon="pi pi-times"
                        variant="text"
                        severity="danger"
                        size="small"
                        :rounded="true"
                        :aria-label="$gettext('Cancel')"
                        :disabled="isSavingAttribution"
                        @click="cancelEditingAttribution"
                    />
                </div>
```

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/SchemeAttributionField.vue
```diff
@@ -0,0 +1,171 @@
...
+}
+
+.attribution-input-wrapper {
+    flex-basis: 100%;
+    min-width: 0;
+}
+
+.attribution-input-wrapper :deep(textarea) {
+    width: 100%;
+}
```
johnatawnclementawn, line 174:
```suggestion
.attribution-input-wrapper {
    display: flex;
    flex-direction: row;
    flex-basis: 100%;
    min-width: 0;
}

.attribution-input-wrapper :deep(textarea) {
    width: 100%;
}

.attribution-edit-controls {
    display: flex;
    flex-direction: column;
}
```

johnatawnclementawn (changes requested):
One last thing - the UX is not consistent with other fields in the latest change. The suggested changes push the controls to the right of the text area & fix tab order.

arches_lingo/src/arches_lingo/components/scheme/SchemeHeader/components/SchemeAttributionField.vue
chrabyrd: You're right, I thought the buttons would be better in a separate line but that _does_ break pattern. undoing!

johnatawnclementawn (approved):
lgtm 👍🏼

PR #753, opened 2026-06-05
adds ability to lock/unlock schemes and related concepts

chrabyrd:


arches_lingo/views/api/scheme_lock.py
```diff
@@ -0,0 +1,79 @@
...
+
+from arches_lingo.mixins.permissions import LingoAdminMixin
+from arches_lingo.utils.concept_lifecycle import (
+    EDITING_STATE_ID,
+    LOCKED_STATE_ID,
+    RETIRED_STATE_ID,
+)
+
+
+class SchemeLockView(LingoAdminMixin, View):
```
johnatawnclementawn, line 18: I won't block on this, but it seems like two views to bit flip + run side effects during `save` seems duplicative. Could this be simplified?

arches_lingo/views/api/scheme_lock.py
```diff
@@ -0,0 +1,79 @@
...
+        if scheme.resource_instance_lifecycle_state_id != LOCKED_STATE_ID:
+            return JSONErrorResponse(
+                title=_("Not locked"),
+                message=_("This scheme is not currently locked."),
+                status=HTTPStatus.BAD_REQUEST,
+            )
+
+        current_state = scheme.resource_instance_lifecycle_state
+        scheme.resource_instance_lifecycle_state = (
+            ResourceInstanceLifecycleState.objects.get(pk=EDITING_STATE_ID)
```
johnatawnclementawn, line 72: Should `Published` schemes be able to be locked? When unlocking a scheme that's been published, then locked, the lifecycle state always returns to edit. I'd expect it to return to `Published`.

Lifecycle timeline: `Published` -> `Locked` -> `Unlock` -> ~~`Edit`~~ `Published`

arches_lingo/utils/scheme_lock.py
```diff
@@ -0,0 +1,35 @@
...
+        pk=scheme_pk,
+        resource_instance_lifecycle_state_id=LOCKED_STATE_ID,
+    ).exists()
+
+
+def get_scheme_id_for_concept(concept_pk: str) -> str | None:
+    tile = TileModel.objects.filter(
+        resourceinstance_id=concept_pk,
+        nodegroup_id__in=[
+            TOP_CONCEPT_OF_NODE_AND_NODEGROUP,
```
johnatawnclementawn: nit - not a big deal, but a concept should always have `CONCEPTS_PART_OF_SCHEME_NODEGROUP_ID`, why not just filter on that nodegroup alone?

johnatawnclementawn (commented):
looks good - a few comments for your consideration

arches_lingo/views/api/scheme_lock.py
chrabyrd: Yeah the intention is to have both states be lockable. Though I hear you on oddness of always reverting to the `editing` state. Unfortunately I think it's out-of-scope to have this thing be able to revert to whichever state it was locked from, but definitely happy to double-check about desired behavior here 👍

arches_lingo/views/api/scheme_lock.py
chrabyrd: I hear you but I prefer the intentionality of keeping them separate. I'd like to leave this as-is unless you feel strongly 👍

arches_lingo/utils/scheme_lock.py
chrabyrd: I don't believe that top concepts have a CONCEPTS_PART_OF_SCHEME_NODEGROUP_ID tile; they only have a TOP_CONCEPT_OF_NODE_AND_NODEGROUP tile

arches_lingo/views/api/scheme_lock.py
johnatawnclementawn: Nah, I don't feel that strongly about it 👍🏼

arches_lingo/views/api/scheme_lock.py
johnatawnclementawn: I get that, we'd have to store the history of the lifecycle state rather than just updating vals... idk, maybe @robgaston has something more to add here, but I think this is fine for now

arches_lingo/utils/scheme_lock.py
johnatawnclementawn: Sorry, I think that's wrong: here's two instances of top concepts, one created through the UI & one imported from a skos file

Screenshot 2026-06-16 at 5 03 12 PM

Screenshot 2026-06-16 at 5 03 41 PM

chrabyrd: OH SNAP! you're right! 🚀

johnatawnclementawn (changes requested):
Can you fix the migration naming collision? 

Side note: should we adopt Core's migration naming convention `[ticket #]_[description]` rather than an `[i++]_[description]`?

chrabyrd:
> Can you fix the migration naming collision?
> 
> Side note: should we adopt Core's migration naming convention `[ticket #]_[description]` rather than an `[i++]_[description]`?

done!

And a side response: I'm down, from a larger perspective. Getting a pattern for things like this out is important. I didn't want to break the mold we've set in Lingo so far with this. Adding on, I don't _love_ the migration naming pattern in arches core but changing it is a larger conversation. the current `[ticket #]_[description]` can be counter-intuitive when spelunking as these things can get out of order depending on when branches are merged. Personally I'd prefer `[i++]_[ticket #]_[ticket name]` but yeah larger conversation 👍

johnatawnclementawn (approved):
LGTM!

PR #754, opened 2026-07-14
Puts exported CSVs into folders to prevent clashes

robgaston:
CSV files are named per nodegroup, but this results in clashes between Concepts and Schemes (for example, both have a Label nodegroup).  To handle this, CSVs are nested within folders for each model.

### DIFF
--- a/arches_lingo/etl_modules/lingo_resource_exporter.py
+++ b/arches_lingo/etl_modules/lingo_resource_exporter.py
@@ -317,25 +317,29 @@ class LingoResourceExporter:
         """
         Export scheme and concept resources as Arches TileCSV format using the
         core TileCsvWriter. Each nodegroup card becomes a separate CSV file,
-        all bundled together in the output ZIP.
+        all bundled together in the output ZIP. Scheme and concept graphs share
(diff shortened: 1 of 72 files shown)


arches_lingo/etl_modules/lingo_resource_exporter.py
```diff
@@ -317,25 +317,29 @@ def _export_as_tilecsv(self, scheme_ids, concept_ids):
...
+            ):
+                content = exported_file["outputfile"].getvalue()
                 if isinstance(content, str):
                     content = content.encode("utf-8")
-                output_files.append({"name": f["name"], "content": content})
+                output_files.append(
+                    {"name": f"{folder}/{exported_file['name']}", "content": content}
+                )
+
+        if scheme_ids:
```
johnatawnclementawn, line 338: This is likely out of scope for this PR, for the sake of getting the Lingo 1.1.0 release out, but we should also handle for the `Group`, `Person`, `Digital Object`, and `Textual Work` models as they also have cards that share names

johnatawnclementawn (approved):
lgtm 👍🏼  - comment can be converted into an issue to target the next release.

PR #755, opened 2026-07-22
adds scripts to dump and load fixtures

robgaston:
these scripts account for images associated with concepts and improve performance for large datasets

### DIFF
--- a/arches_lingo/utils/lingo_fixtures.py
+++ b/arches_lingo/utils/lingo_fixtures.py
@@ -50,20 +50,21 @@ from django.core.files.storage import default_storage
 from django.core.management import call_command
 from django.db import connection, transaction
 
-from arches.app.models.models import File, GraphModel
+from arches.app.models.models import (
+    File,
+    GraphModel,
+    ResourceInstance,
+    ResourceIdentifier,
+    ResourceXResource,
+    TileModel,
+)
 
-# Every resource graph shipped with arches-lingo that a concept or scheme tile
-# can reference: the concept/scheme resources themselves, the actor and
-# citation resources used to record attribution, and the digital-object
-# resources used for concept images.
-LINGO_FIXTURE_GRAPH_SLUGS = [
-    "concept",
-    "scheme",
-    "digital_object_system",
-    "person_system",
-    "group",
-    "textual_work",
-]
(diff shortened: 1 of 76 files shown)


arches_lingo/utils/lingo_fixtures.py
```diff
@@ -0,0 +1,319 @@
...
+# citation resources used to record attribution, and the digital-object
+# resources used for concept images.
+LINGO_FIXTURE_GRAPH_SLUGS = [
+    "concept",
+    "scheme",
+    "digital_object_system",
+    "person_system",
+    "group",
+    "textual_work",
+]
```
johnatawnclementawn: Can we move this to `const.py`?

arches_lingo/utils/lingo_fixtures.py
```diff
@@ -0,0 +1,319 @@
...
+    where_sql: str
+    # Bigint identity-column tables need their sequence bumped after a COPY
+    # restores explicit id values, or the next ORM-created row will collide.
+    identity_column: str | None = None
+
+
+FIXTURE_TABLES: tuple[FixtureTable, ...] = (
+    FixtureTable(
+        name="resource_instances",
+        columns=(
```
johnatawnclementawn: I totally dig the performance improvement by moving from ORM-based django fixture creation -> SQL-based fixture creation, but one thing the former gets right is always staying in sync with the state of the models at runtime. Not that we have a bunch of churn in our models, but I think it would be better to derive the columns direct from the model classes rather than hard-coding.

johnatawnclementawn (commented):
Looks good and works as expected - two notes for your consideration.

Also, any reason for splitting commands into separate files?

arches_lingo/utils/lingo_fixtures.py
robgaston: good call, i reworked `FixtureTable` to derive `name` and `columns` from a specified `model`

robgaston:
> Also, any reason for splitting commands into separate files?

I just did that cause having these used as separate subcommands seemed like a better design (with shared logic in the util), but happy to reconsider.

johnatawnclementawn (approved):
lgtm 👍🏼

PR #756, opened 2026-07-22
adds advanced search facet for searching on related images

robgaston:


arches_lingo/utils/advanced_search.py
```diff
@@ -52,6 +52,15 @@
...
     CONCEPT_TYPE_NODEID,
+    DEPICTING_DIGITAL_ASSET_INTERNAL_NODEGROUP,
+    DEPICTING_DIGITAL_ASSET_INTERNAL_NODE,
+    DEPICTING_DIGITAL_ASSET_EXTERNAL_NODEGROUP,
+    DIGITAL_OBJECT_CONTENT_NODEGROUP,
+    DIGITAL_OBJECT_CONTENT_NODE,
+    DIGITAL_OBJECT_NAME_NODEGROUP,
+    DIGITAL_OBJECT_NAME_CONTENT_NODE,
+    DIGITAL_OBJECT_STATEMENT_NODEGROUP,
+    DIGITAL_OBJECT_STATEMENT_CONTENT_NODE,
```
chrabyrd, line 63: do we need to add indices like in `0012` or `0014`?

arches_lingo/utils/advanced_search.py
```diff
@@ -654,6 +664,151 @@ def _facet_attribution_source(self, condition):
...
 
+    def _concepts_depicting(self, digital_object_ids):
+        """Return concept PKs whose internal depiction references any digital object.
+
+        Matches the ``depicting_digital_asset_internal`` resource-instance-list node
+        against the given digital-object resource IDs via JSONB containment.
+        """
+        containment = Q()
+        matched_any = False
+        for digital_object_id in digital_object_ids:
```
chrabyrd: looping on digital object ids and OR-ing them together? If this gets large it won't be super performant. rawsql with an exists like in `_digital_objects_by_file_name` would scale better 👍

arches_lingo/utils/advanced_search.py
```diff
@@ -654,6 +664,151 @@ def _facet_attribution_source(self, condition):
...
+            )
+        else:
+            if match_mode == "exact":
+                pattern = value
+            elif match_mode == "starts_with":
+                pattern = f"{value}%"
+            elif match_mode == "ends_with":
+                pattern = f"%{value}"
+            else:  # contains
+                pattern = f"%{value}%"
```
chrabyrd: would value need to be escaped here?

chrabyrd (changes requested):
overall looks good! just a few Qs

arches_lingo/utils/advanced_search.py
robgaston: Good call; i added migration `0018_add_related_image_gin_indexes` with trigram GIN indexes on the Digital Object title and description text nodes

arches_lingo/utils/advanced_search.py
robgaston: done! 👍

arches_lingo/utils/advanced_search.py
robgaston: yeah, you're right. i added a helper method (`_escape_like_wildcards`) for this

chrabyrd (approved):
LGTM!

PR #759, opened 2026-07-24
Adds content-negotiated SKOS dereferencing for scheme & concept URIs

robgaston:
serves scheme/concept identity URIs as either HTML or SKOS based on the Accept header or a ?format= override, supporting RDF/XML, Turtle, N-Triples, and JSON-LD. published resources are dereferenceable anonymously; drafts require authentication and retired concepts are marked owl:deprecated.

also, uses each resource's minted public URI as the RDF subject (with owl:sameAs back to the UUID URI), serialize skos mapping properties (exactMatch/closeMatch/etc.) as URI references, and enumerate skos:hasTopConcept when serializing a scheme alone.

### DIFF
--- a/arches_lingo/utils/concept_lifecycle.py
+++ b/arches_lingo/utils/concept_lifecycle.py
@@ -16,11 +16,11 @@ PUBLISHED_STATE_ID = uuid.UUID("6b0f1a7b-5b3d-4b2a-8a5b-7c3a1b0f2d9e")
 RETIRED_STATE_ID = uuid.UUID("9d2e1c0b-7a6b-4b3d-8c1a-0f2d9e6b0a7c")
 LOCKED_STATE_ID = uuid.UUID("c9f7e3d1-2a5b-4c8d-9e6f-3b0a1d2e4f7c")
 
-# Lifecycle states in which a resource has been published and its URI is a stable,
-# publicly resolvable identifier. Draft and Editing resources are pre-publication and
-# are not dereferenceable to anonymous consumers. Retired concepts remain resolvable so
-# that previously published URIs do not break; they are marked as deprecated in the
-# serialized output.
+# Lifecycle states whose URIs we treat as publicly resolvable to anonymous consumers.
+# Retired resources stay resolvable (marked owl:deprecated in the serialized output) so
+# that previously published URIs do not break. These states do not strictly guarantee
+# prior publication: a resource can be retired or locked directly from draft, in which
+# case it is anonymously resolvable here as well.
 PUBLICLY_DEREFERENCEABLE_STATE_IDS = frozenset(
     {PUBLISHED_STATE_ID, LOCKED_STATE_ID, RETIRED_STATE_ID}
 )
--- a/arches_lingo/utils/concept_lifecycle.py
+++ b/arches_lingo/utils/concept_lifecycle.py
@@ -12,8 +12,18 @@ from arches_lingo.const import (
 
 DRAFT_STATE_ID = uuid.UUID("0e7f8c6d-1f7b-4c2a-9a0c-2b9e0d6c8f11")
 EDITING_STATE_ID = uuid.UUID("b3a6a0d2-2b5c-4c2f-9d6c-0c2a5b7d1e8f")
+PUBLISHED_STATE_ID = uuid.UUID("6b0f1a7b-5b3d-4b2a-8a5b-7c3a1b0f2d9e")
 RETIRED_STATE_ID = uuid.UUID("9d2e1c0b-7a6b-4b3d-8c1a-0f2d9e6b0a7c")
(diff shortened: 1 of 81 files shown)


arches_lingo/utils/skos_serializer.py
```diff
@@ -0,0 +1,397 @@
...
+        "owl": str(OWL),
+    }
+    compacted = jsonld.compact(expanded, context)
+    return json.dumps(compacted, indent=2, sort_keys=True).encode("utf-8")
+
+
+# --- Content negotiation -------------------------------------------------------
+
+# Registry of the RDF serializations Lingo exposes for dereferenced resources.
+# Keyed by the short format token also accepted as a ?format= query value.
```
chrabyrd: nit throughout PR: superflous code comments

arches_lingo/utils/concept_lifecycle.py
```diff
@@ -12,9 +12,19 @@
...
 LOCKED_STATE_ID = uuid.UUID("c9f7e3d1-2a5b-4c8d-9e6f-3b0a1d2e4f7c")
 
+# Lifecycle states in which a resource has been published and its URI is a stable,
+# publicly resolvable identifier. Draft and Editing resources are pre-publication and
+# are not dereferenceable to anonymous consumers. Retired concepts remain resolvable so
+# that previously published URIs do not break; they are marked as deprecated in the
+# serialized output.
+PUBLICLY_DEREFERENCEABLE_STATE_IDS = frozenset(
+    {PUBLISHED_STATE_ID, LOCKED_STATE_ID, RETIRED_STATE_ID}
+)
```
chrabyrd: This assumption isn't true, it's possible to both `retire` and `lock` schemes/concepts from the draft state, without first being published.

arches_lingo/utils/skos_serializer.py
```diff
@@ -0,0 +1,397 @@
...
+RDF_FORMATS = {
```
chrabyrd: declaring constants mid-file?

arches_lingo/views/api/dereference.py
```diff
@@ -0,0 +1,121 @@
...
+        lifecycle_state_id = self._lifecycle_state_id(resource_id)
+        if not self._may_dereference(request, lifecycle_state_id):
+            raise Http404()
+
+        body, content_type = serialize_resource(
+            resource_id,
+            graph_kind,
+            rdf_format,
+            mark_deprecated=lifecycle_state_id == RETIRED_STATE_ID,
+        )
```
chrabyrd: this should probably be wrapped in a try/except so the api isn't surfacing an unhandled 500 if serialization fails

arches_lingo/utils/skos.py
```diff
@@ -386,8 +415,13 @@ def extract_predicate_object(self, triple):
         object_language = triple.get("object_language")
         if object_language:
             object = Literal(object, lang=object_language)
+        elif triple.get("object_is_uri") and object:
+            # A matched concept's comparate is a URI reference, not a literal.
+            object = URIRef(str(object))
```
chrabyrd: an empty comparate falls through `None` here I think

chrabyrd: also do we need other validation here? this would pass with a non-uri string

chrabyrd (changes requested):
overall looks good and useful! as long as we can either prune lifecycle buttons or otherwise navigate around _that_ weirdness this should be g2g 👍

arches_lingo/utils/skos_serializer.py
robgaston: good point; i tried to clean these up

arches_lingo/utils/concept_lifecycle.py
robgaston: yeah, you're right. I kept the set as-is (we want retired URIs to stay resolvable + owl:deprecated) but fixed the comment to call out the draft->retired/lock edge explicitly. I think we can live with this edge case for now, but happy to revisit

arches_lingo/utils/skos_serializer.py
robgaston: fixed!

arches_lingo/views/api/dereference.py
robgaston: i wrapped `serialize_resource` in a try/except; it now logs and returns a 500 instead of surfacing a traceback

arches_lingo/utils/skos.py
robgaston: good catch; I added `uri_reference_or_none` helper function to handle this

robgaston: this is also handled by `uri_reference_or_none` now

chrabyrd (approved):
LGTM!

PR #766, opened 2026-08-04
ensures that lingo editors can export without additional groups

robgaston:


arches_lingo/src/arches_lingo/api.ts
```diff
@@ -737,20 +737,19 @@ export const exportThesaurus = async (
...
     if (fileName) {
         formData.append("filename", fileName);
     }
-    const response = await fetch(arches.urls.etl_manager, {
-        method: "POST",
-        headers: { "X-CSRFTOKEN": getToken() },
-        body: formData,
-    });
+    const response = await fetch(
+        generateArchesURL("arches_lingo:api-lingo-export"),
```
chrabyrd, line 746: <3

